PR #11, opened 2020-10-25
Suvelak_Assignment_5

SuvelaKoct2020:
-> added for practice

### DIFF
--- a/src/suvela/Mall.java
+++ b/src/suvela/Mall.java
@@ -1,0 +1,61 @@
+package suvela;
+
+class Mall {
+	int items;
+	String cust_Name;
+	String visa_card;
+	String cash;
+	String payment;
+	boolean bill;
+	boolean carry_Bag;
+	
+	
+	void itemAddedToCart(int items) {
+			this.items = items;
+	}
+	void customerName(String cust_Name) {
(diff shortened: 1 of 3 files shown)


src/suvela/Mall.java
```diff
@@ -0,0 +1,61 @@
...
+	boolean carry_Bag;
+	
+	
+	void itemAddedToCart(int items) {
+			this.items = items;
+	}
+	void customerName(String cust_Name) {
+			this.cust_Name = cust_Name;
+	}
+	void paymentDoneUsing(String visa_card, String cash) {
```
harshvegada, line 19: why we make this method as two parameterized, we need only one payment method

harshvegada: variable name should be paymentType not visa_card & cash

harshvegada (commented):
2nd Program is missing,

Need code improvement!
Class is not up to the mark, 
Requirement not fulfilled the way it should fulfill.

Refer this code & compare: https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/23/files

harshvegada (approved):
Code As per the exaptation's, Keep trying your best! @SuvelaKoct2020

PR #13, opened 2020-10-25
PrasadC_Assignment_5

PrasadCOct2020:
>>Wrote a class which accepts Car manufacturer name and displays available model names .  

Please review.

Thanks!

### DIFF
--- a/src/prasad/StudentMarks.java
+++ b/src/prasad/StudentMarks.java
@@ -1,0 +1,38 @@
+package prasad;
+
+public class StudentMarks {
+	String studentName;
+	int marks;
+	Boolean flag;
+	
+	void studentName(String name , int marks) {
+		studentName = name;
+		this.marks = marks;
+	}
+	
+	void gainedMarks() {
+		if(marks>=90) {
+			flag=true;
+			System.out.println(studentName + " has secured A+ grade.");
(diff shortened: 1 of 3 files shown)


PrasadCOct2020:
@KrishnaKTechnocredits I later on realized that in mall class I should have used **<=10**  instead of **<10** in void decideSection() method's first If block to handle use case when customer has added 10 items.

src/prasad/StudentMarks.java
```diff
@@ -0,0 +1,38 @@
...
+		}else if(marks<60) {
+			flag=true;
+			System.out.println(studentName + " has secured C grade.");
+		}if(flag)
+			 System.out.println("Student result was processed successfully.");
+	}
+	
+	public static void main (String[] arg) {
+		StudentMarks studentmarks = new StudentMarks();
+		studentmarks.studentName("Alan", 92);
```
harshvegada, line 35: You written method name as studentName() and you are passing two parameter into method, method name should high level idea about implementation.

harshvegada (approved):
Hi @PrasadCOct2020 
Code should be as per requirement(method name), so it will be easy to identify.()no logical impact), Code looks correct logically, please work on comments which given.

PR #14, opened 2020-10-25
ApurvaC_Assignment_1

ApurvaQ:
Added for test

### DIFF
--- a/src/apurva/Mall.java
+++ b/src/apurva/Mall.java
@@ -1,50 +1,65 @@
 package apurva;
 
 class Mall {
+
+	String custName;
+	int items;
+	boolean bill;
+	boolean carryBag;
+	String payment;
+	
 	void itemAddedToCart(int items) {
-		System.out.println(items);
+		this.items = items;
 	}
 
-	void customerName(String name) {
-		System.out.println(name);
+	void customerName(String custname) {
+		this.custName = custname;
 	}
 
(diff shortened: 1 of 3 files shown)


src/apurva/Mall.java
```diff
@@ -0,0 +1,50 @@
+package apurva;
+
+class Mall {
+	void itemAddedToCart(int items) {
+		System.out.println(items);
```
harshvegada, line 5: why we are printing statement inside method, this are method is not for display items,

src/apurva/Mall.java
```diff
@@ -0,0 +1,50 @@
...
+		else if (items>10 && payment==visaCard)
+			System.out.println();
+		else if (items>10 && payment==cash)
+			System.out.println();
+		else if (items<10 && payment==visaCard)
+			System.out.println();
+	}
+
+	void needBill(boolean bill) {
+		System.out.println(true);
```
harshvegada, line 29: Here, by printing true, how customer will know about true! same for haveCarryBag()

harshvegada (commented):
Hi @ApurvaQ 
Need to work on First Program, looks like you are not clear on requirement, refer below link.
Second Programs as per the expectations

refer: https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/23/files
for first program

harshvegada (approved):
Code as per the expectations, Please give correct assignment number.

PR #15, opened 2020-10-25
NagendraT_Assignment_5

NagendraTOct2020:
-->NagendraT_Assignment_5

### DIFF
--- a/src/nagendra/Mall.java
+++ b/src/nagendra/Mall.java
@@ -1,0 +1,60 @@
+package nagendra;
+
+public class Mall {
+
+	int numberOfItems;
+	String customername;
+	String paymentType;
+	String section;
+	boolean isBillRequired;
+	boolean doesCustomerNeedCarryBag;
+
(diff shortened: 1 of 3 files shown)


src/nagendra/Mall.java
```diff
@@ -0,0 +1,60 @@
...
+
+	void decideSection() {
+		if (numberOfItems < 10 && paymentType.equals("cash")) {
+			section = "section 1A";
+		} else if (numberOfItems > 10 && paymentType.equals("visa card")) {
+			section = "section 1B";
+		} else if (numberOfItems > 10 && paymentType.equals("cash")) {
+			section = "section 1c";
+
+		} else {
```
harshvegada, line 45: You should add condition for visa card(4th condition), "else" will execute always if customer passing other value than visa & cash, we need to add that 4th condition.
This is also fine for now, my point is to you know other scenarios

harshvegada (approved):
Good job @NagendraTOct2020 
Code looks Good to me!
Keep it up!

PR #18, opened 2020-10-25
SurbhiJ_Assignment_5

SurbhiJOct20:
----->Added for practice


harshvegada (approved):
Hi @SurbhiJOct20 
Very well done code!
One suggestion here for Student class, we should display grade while calling grade() student should not call display() to check their grade. you need to make grade method as parameterized.

Logically Code looks good to me!
Keep it up!
Keep trying your best!

PR #6, opened 2020-10-25
VaibhavJ_Assignment_5

VaibhavJOct2020:
-- Added for Practice

PR #20, opened 2020-10-25
BrijeshG_Assignment_3

BrijeshGOct2020:


src/brijesh/Mall.java
```diff
@@ -0,0 +1,66 @@
...
+	void customerName(String name) {
+		this.name=name;
+	}
+	
+	void paymentDoneUsing(String paymentMode) {
+		this.paymentMode=paymentMode;	
+	}
+	
+	void displayCutomerInfo(Mall mall) {
+		mall.decideSection();
```
harshvegada, line 24: Here, you can directly call the decideSection() method.
this is also correct. 
you should call decideSection()  from the main method not from any internal method() bit different than requirement.
Refer: https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/6/files

harshvegada (approved):
Good job @BrijeshGOct2020 
Class logically correct!
bit different than requirement, you are calling decideSection() from internal method. it should be called from main method only. please work on comment in next PR

PR #28, opened 2020-10-26
Anjali p asgmt 5

AnjaliPOct2020:
Assignment 5 > Mall Scenario and Student Marks task created using if else and else if ladder method.


harshvegada (approved):
Hi @AnjaliPOct2020,
One suggestion here for Student class, we should display grade while calling grade() student should not call display() to check their grade. you need to make grade method as parameterized.


Logically Code looks good to me!
Keep it up!
Keep trying your best!

PR #35, opened 2020-10-26
Anurag_Assignment5

anuragkoct2020:
->Added file Student.java
->Added file Mall.java 
[under package  'Anurag']

Please review my PR.

### DIFF
--- a/src/anurag/Mall.java
+++ b/src/anurag/Mall.java
@@ -1,0 +1,55 @@
+package anurag;
+
+public class Mall {
+	int item;
+	String customName;
+	String visaCard ;
+	String cash;
+    String section;
+    String paymentMethod;
+    
+    void itemAddedToCart(int item) {
+		this.item = item;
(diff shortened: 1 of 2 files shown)


src/anurag/Mall.java
```diff
@@ -0,0 +1,55 @@
...
+		this.item = item;
+		}
+	void customerName(String customName ) {
+		this.customName=customName;
+		}
+	void paymentDoneUsing(String paymentMethod) {
+		this.paymentMethod = paymentMethod;
+		}
+	void needBill(boolean bill){ 
+		 if(bill==true)
```
harshvegada, line 21: here you can write only bill only no need to compare with true, if **bill** is true then it will comes under if condition. same for haveCarryBag()

harshvegada (approved):
Code is as per the requirement @anuragkoct2020 
Keep it up!
Good to merge!

PR #36, opened 2020-10-26
RitikaG_Assignment_5

RitikaGOct2020:
--> Added for practice purpose

### DIFF
--- a/src/ritika/Mall.java
+++ b/src/ritika/Mall.java
@@ -1,0 +1,93 @@
+/* 1. Create a class named as Mall, which contains below methods
+a. itemAddedToCart()
+b. customerName()
+c. paymentDoneUsing()
+	i. Visa Card 
+	ii. Cash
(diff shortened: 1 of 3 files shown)


src/ritika/Mall.java
```diff
@@ -0,0 +1,93 @@
...
+		else if (carryBag == false)
+			System.out.println(customerName + " does not have Carry Bag");
+	}
+	
+	void itemAddedToCart(int numberOfItems){
+		this.numberOfItems = numberOfItems;
+	}
+	
+	void paymentDoneUsingCardOrCash(String paymentMethod) {
+		if (paymentMethod.equals("Card"))
```
harshvegada, line 59: instead of using true & false, you should use only String
this.paymentMethod = paymentMethod

harshvegada (approved):
Code looks good, keep it up!
improve code, decide which data type you will use.

PR #38, opened 2020-10-26
MonikaS_Asgmt_5

MonikaSOct2020:
Assignment 5 student and Mall program

### DIFF
--- a/src/monika/Student.java
+++ b/src/monika/Student.java
@@ -1,0 +1,71 @@
+package monika;
+class Student{
+
+    String nameOfStudent;
+	int marksGained;
+	String grade;
+  
+    void studentName(String nameOfStudent){
+	   this.nameOfStudent = nameOfStudent;
+	  
+	}
+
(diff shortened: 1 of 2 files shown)


src/monika/Student.java
```diff
@@ -0,0 +1,71 @@
...
+	
+	   	System.out.println(nameOfStudent+" secured grade "+grade+ " in the exam");
+	   
+	}
+
+   public static void main(String[] args) {
+     Student student = new Student();
+	 student.studentName("john");
+	 student.gainedMark(90);
+	 student.studentCritera();
```
harshvegada, line 51: Here, ideally student should not call studentCritera() method to check result/grade, while passing marks we should display the grade of student.

Keep the for now, in next PR improve this point.

harshvegada (approved):
Code looks Good to me, as per the requirement coded, @MonikaSOct2020 
Work on comments, rest logically both class good
Keep it up!
Good to merge!

PR #39, opened 2020-10-27
Sadhana p asgmt 5

SadhanaPOct2020:


harshvegada (approved):
You should display name while calling gainedMark(), why student should call studnetMarkDetails() method to check grade, rest code good.

Please follow best practices on variable & method

PR #43, opened 2020-10-27
PrasadC_Assignment6

PrasadCOct2020:
Please review my PR.

Thanks!

### DIFF
--- a/src/prasad/Shop.java
+++ b/src/prasad/Shop.java
@@ -1,0 +1,62 @@
+package prasad;
+
+public class Shop {
+	int maggie = 50;
+	int oilPouch = 0;
+	int paniPuri = 43;
+	int masalaPackets = 73;
+
+	void customerorder(int item1,int item2,int item3,int item4) {
+		if(maggie==0)
+			System.out.println("[!] Sorry Maggie is running out of stock."); 
+		else if (item1>maggie) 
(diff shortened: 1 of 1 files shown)


src/prasad/Shop.java
```diff
@@ -0,0 +1,62 @@
...
+	int oilPouch = 0;
+	int paniPuri = 43;
+	int masalaPackets = 73;
+
+	void customerorder(int item1,int item2,int item3,int item4) {
+		if(maggie==0)
+			System.out.println("[!] Sorry Maggie is running out of stock."); 
+		else if (item1>maggie) 
+			System.out.println("Sorry quantity ordered for maggie cannot be fulfilled.");
+			else maggie = maggie-item1;
```
harshvegada, line 14: Here, You should display available stock for Maggie. with in else if() condition
remove else statement from line 14th

same for others

harshvegada (commented):
Work on given comment, rest logic looks good to me!

src/prasad/Shop.java
PrasadCOct2020: Hi @harshvegada I have included additional condition to check if customer has ordered more items than what are available in store and then convey the user with an error that stated quantity cannot be fulfilled. I have written separate method to display items available in inventory post accepting item counts from user. I have tested the code and it works fine.

harshvegada (approved):
got it!

PR #45, opened 2020-10-27
VishakhaB_Assignment_6

VishakhaBOct2020:
Please review my PR for assignment 6.


harshvegada (approved):
Excellent work @VishakhaBOct2020 
One suggestion if you have stock then write down stocks behind that variable which you did, for 1-2 may be missed that's fine.
Impressive work!
Good to merge

PR #49, opened 2020-10-27
Milind_Assignment_6

MIlindChavhanOct2020:
MilindC_Assignment_6

### DIFF
--- a/src/milind/Shop.java
+++ b/src/milind/Shop.java
@@ -1,0 +1,106 @@
+package milind;
+
+public class Shop {
+	
+	static int stocksOfMaggie = 50;
+	static int stocksOfDosa = 43; 
+	static int stocksOfPouches = 39; 
+	static int stocksOfPanipuri = 43;  
+	static int stocksOfMasala = 73;
+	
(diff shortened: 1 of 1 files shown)


src/milind/Shop.java
```diff
@@ -0,0 +1,106 @@
...
+	static int stocksOfPouches = 39; 
+	static int stocksOfPanipuri = 43;  
+	static int stocksOfMasala = 73;
+	
+	void customerOrder(int orderOfMaggy, int orderOfDosa, int orderOfPouches, int orderOfPanipuri, int orderOfMasala) {
+		
+		if(orderOfMaggy <= stocksOfMaggie) {
+			stocksOfMaggie = stocksOfMaggie - orderOfMaggy;
+		} else {
+			System.out.println("Maggy packets are running out of stock");
```
harshvegada, line 16: If we have 50 packet and customer asking for 51 packets, then we should not display message like we are out of stock! you should display remain packet so that will clear customer, atlases customer can buy this much packets. same for all other methods! some where logically looks wrong this approach, think on this. 

Do not change now, keep as it is.
improve this in next PR

harshvegada (commented):
accommodate change @MIlindChavhanOct2020

harshvegada (approved):
Hi @MIlindChavhanOct2020 
Code Looks Good to me!
Good to merge

PR #53, opened 2020-10-27
AnjaliP_Assignment_6

AnjaliPOct2020:
Assignment 6 for Shop Scenario Created

### DIFF
--- a/src/anjaliP/Shop.java
+++ b/src/anjaliP/Shop.java
@@ -1,0 +1,84 @@
+package anjaliP;
+
+public class Shop {
+	static int maggie = 50;
+	static int dosa = 43;
+	static int oilPouch = 39;
+	static int paniPuri = 43;
+	static int masala = 73;
+	
+	void customerItems(int maggie1 , int dosa1 , int oilPouch1 , int paniPuri1 , int masala1) 
+	{		
+		System.out.println("Customer Purchased :- "+"\n"+"--------------------------------------"
+				+"\n"+maggie1+" Maggie Packets"+"\n"+dosa1+" Dosa packets"+"\n"+oilPouch1+
+				" Oil Packets"+"\n"+paniPuri1+" Pani Puri packets "+"\n"+masala1+" Masala Packets. ");
+		System.out.println("----------------------Current status of Items in stock are----------------");
+		if (maggie==0)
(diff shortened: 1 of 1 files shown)


src/anjaliP/Shop.java
```diff
@@ -0,0 +1,84 @@
...
+	else if (oilPouch!=0)
+		System.out.println("OilPouch Available in Stock "+oilPouch);
+	else if (paniPuri!=0)
+		System.out.println("PaniPuri Available in Stock "+paniPuri);
+	else if (masala!=0)
+		System.out.println("Masala packet Available in Stock "+masala);
+	}
+	public static void main(String[] a) {
+		Shop shop = new Shop();
+		shop.customerItems(25, 30, 17, 25, 43);
```
harshvegada, line 79: If we have 25 packet and customer asking for 26 packets, then we should not display message like we are out of stock! you should display remain packet so that will clear customer, atlases customer can buy this much packets. same for all other methods! some where logically looks wrong this approach, think on this.

src/anjaliP/Shop.java
```diff
@@ -0,0 +1,84 @@
...
+		System.out.println("Customer Purchased :- "+"\n"+"--------------------------------------"
+				+"\n"+maggie1+" Maggie Packets"+"\n"+dosa1+" Dosa packets"+"\n"+oilPouch1+
+				" Oil Packets"+"\n"+paniPuri1+" Pani Puri packets "+"\n"+masala1+" Masala Packets. ");
+		System.out.println("----------------------Current status of Items in stock are----------------");
+		if (maggie==0)
+			System.out.println("Maggie is Out of Stock");
+		else if (maggie1>maggie)
+			System.out.println("Maggie is currently running out of Stock");
+		else maggie = maggie - maggie1;
+			System.out.println("Maggie Available in Stock : "+maggie);
```
harshvegada, line 21: Here, If If we have 25 packet and customer asking for 26 packets, then you are displaying running out of Stock message! which is wrong, 
if this is the case then you should display remaining packets right, which you did, but one mistake you written in else if() condition so any single condition will executed.

think on this.

harshvegada (commented):
work on comments

harshvegada:
Looks like change not added, merging this PR

PR #60, opened 2020-10-27
RaghvendraT_Assignment_6

RaghvendramaniTOct2020:


harshvegada (approved):
Code is as per the requirement, if you make variable as non-static still this gone work! that's fine, but you got idea, what m trying to say.
improve in next PR.
Keep doing well!
Good for merge.

PR #71, opened 2020-10-27
RubyK_Assignment_6

rubykoct2020:
RubyK_Assignment_6


src/ruby/Shop.java
```diff
@@ -0,0 +1,63 @@
...
+package ruby;
+public class Shop {
+	int maggie = 50;
+	int dosa= 43;
+	int oilPouches = 39;
+	int panipuri = 43;
+	int masala = 73;
+	
+	void outOfStock(){
+		if(maggie <= 0)
```
harshvegada, line 24: here, condition should be == not <=0 otherwise quantity never be in minus logically right.
improve in next PR

harshvegada (approved):
Code looks good to me!
Good for merge!

PR #72, opened 2020-10-27
KajolC_Assignment_6

KajolCOct2020:
Assignment 6:
Program-ShopStockDetails

### DIFF
--- a/src/kajol/ShopStockDetails.java
+++ b/src/kajol/ShopStockDetails.java
@@ -73,7 +73,7 @@ public class ShopStockDetails {
 		if(intialStockOfMasala>0)
 			System.out.println("Masala Packets : "+intialStockOfMasala);
 		System.out.println();
-		}
+	}
 
 	public static void main(String[] args) {
 		ShopStockDetails shopStockDetails=new ShopStockDetails();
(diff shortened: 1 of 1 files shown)


src/kajol/ShopStockDetails.java
```diff
@@ -0,0 +1,86 @@
...
+public class ShopStockDetails {
+	static int intialStockOfMaggie=50;
+	static int intialStockODosa=43;
+	static int intialStockOfPouches=39;
+	static int intialStockOfPanipuri=43;
+	static int intialStockOfMasala=73;
+	
+	void itemsToBePurchased(int purchasedMaggie,int purchasedDosa,int purchasedPouches,int purchasedPanipuri,int purchasedMasala) {
+		System.out.println("Items Purchased....");
+		if(purchasedMaggie>intialStockOfMaggie)
```
harshvegada, line 12: If I passpurchasedMaggie as 51, then it will goes into else part & make intialStockOfMaggie in -1, which looks wrong logically,

same for all other methods.

harshvegada (approved):
Code is as per the requirement,
Keep it up!

PR #78, opened 2020-10-27
MonikaS_Asgmt_6

MonikaSOct2020:


src/monika/Shop.java
```diff
@@ -0,0 +1,106 @@
...
+			System.out.println("Panipuri Out of stock");
+		if(masalaStock<=0)
+			System.out.println("Masala Out of stock");
+	}
+	
+	void showInStockItem(){
+		   System.out.println("--In stock--");
+		   
+		   if(maggieStock>0)
+			System.out.println("Maggie in stock");
```
harshvegada, line 68: You should display how many quantity we have left.

src/monika/Shop.java
```diff
@@ -0,0 +1,106 @@
...
+   int pouchesStock=30;
+   int panipuriStock=43;
+   int masalaStock=74;
+   
+   
+   
+   void customerBoughtItems(int maggieCount, int dosaCount, int pouchesCount ,int panipuriCount, int masalaCount){
+       if (maggieStock !=0 && maggieCount<= maggieStock)
+	       maggieStock=maggieStock-maggieCount;
+		   else if (maggieStock<=maggieCount)
```
harshvegada, line 15: If you check your code then second condition in if statement & else if() both are same, it neven goes into  **else if()** condition. same for all other condition.

harshvegada (approved):
code looks good,
work on comments and improve coding in next PR

PR #58, opened 2020-10-27
BrijeshG_Assignment_6

BrijeshGOct2020:

PR #79, opened 2020-10-27
ShrutiC_Assignment_6

ShrutiCOct2020:
--> Created a class Shop and added methods to display the available and out of stock items

### DIFF
--- a/src/shrutiC/Shop.java
+++ b/src/shrutiC/Shop.java
@@ -1,0 +1,99 @@
+package shrutiC;
+
+class Shop {
+
+	static int maggieStock = 50;
+	static int dosaStock = 43;
+	static int oilPouchStock = 39;
+	static int panipuriStock = 43;
+	static int masalaStock = 73;
+
+	void displayOutOfStockItem() {
+
+		System.out.println("\nSorry below items are Out of stock :-");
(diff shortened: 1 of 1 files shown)


src/shrutiC/Shop.java
```diff
@@ -0,0 +1,99 @@
...
+			System.out.println("Masala - " + masalaStock + " packets");
+
+	}
+
+	void orderMaggie(int maggieOrder) {
+
+		if (maggieOrder <= maggieStock) {
+			maggieStock = maggieStock - maggieOrder;
+		} else if (maggieOrder > maggieStock && !(maggieStock == 0)) {
+			maggieStock = 0;
```
harshvegada, line 48: Assume we have 51 maggieOrder then it will go into else if condition, but condition you give bit wrong logically.
(51>50) && !(51==0)
true       &&     true

if condition satisfied then you are making maggieStock  = 0 which wrong.

same for other methods.

harshvegada (commented):
work on comment 
refer this code

https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/58

harshvegada:
Looks like you not update the code as requested. merging PR.

PR #93, opened 2020-10-29
Mrunal_Assignment_07

MrunalYOct2020:


src/mrunal/assignment7.java
```diff
@@ -0,0 +1,59 @@
...
+		
+	}
+	
+	void displayDivisiblebySevenAndThirteen(int start, int end )
+	{
+		for(int i= start; i<=end;i++)
+
+		{
+			if((i%7==0) || (i%13==0))
+			 System.out.println(i +" is divisible by "  );
```
harshvegada, line 38: Here, how u predict that number is divide by 7 or 13. here you need to 
if(number%7==0 )
				System.out.println(number + " is divisible by 7");
			else if(number%13==0)
				System.out.println(number + " is divisible by 13");

harshvegada (approved):
very well @MrunalYOct2020 
Keep doing well!
NO comments.
Good to merge.

harshvegada:
next time take care of this.

PR #101, opened 2020-10-29
SuvelaK_assignment_7

SuvelaKoct2020:


src/suvela/ForLoop.java
```diff
@@ -0,0 +1,44 @@
+package suvela;
+
+public class ForLoop {
+
+	void even(int start, int end) {
+		for (int i = start; i <= end; i++) {
+			System.out.println("Even number is " + i);
```
harshvegada, line 7: Here you should do if(i%2==0) right, consider a scenario where number starts with 11 then your code will not work as expected.

harshvegada (approved):
Rest code looks good to me!
Good to merge!
Keep it up @SuvelaKoct2020

PR #125, opened 2020-10-30
Apurva_Asgmt_7

ApurvaQ:
Please review my code I found an issue, my last method is not print please check once

### DIFF
--- a/src/apurva/NumberPrint.java
+++ b/src/apurva/NumberPrint.java
@@ -1,0 +1,43 @@
+package apurva;
+
+class NumberPrint {
+	void evenNum(int start, int end) {
+		for (int index = start; index <= end; index++) {
+			if (index % 2 == 0) {
+				System.out.println("Even numbers are: " + index);
+			}
+		}
+	}
+
+	void divNum1(int start, int end) {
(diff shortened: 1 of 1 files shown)


src/apurva/NumberPrint.java
```diff
@@ -0,0 +1,43 @@
...
+			if (index % 5 == 0 && index % 3 == 0) {
+				System.out.println("Divisible by 5 & 3 numbers are: " + index);
+			}
+		}
+	}
+
+	void divNum3(int start, int end) {
+		for (int index = start; index <= end; index++) {
+			if (index % 7 == 0 && index % 13 == 0) {
+				System.out.println("Divisible by 7 & 13 numbers are: " + index);
```
harshvegada, line 31: mis-match with code requirement here you need to use || not &&

if (i % 7 == 0)
				System.out.println(i + " is divisible by 7");
			else if (i % 13 == 0)
				System.out.println(i + " is divisible by 13");

take of this in next PR

harshvegada (approved):
work on comments, correct the understanding.
try to avoid this kind mistake.

PR #128, opened 2020-10-30
AnkitJ_Assignment_8

AnkitJOct2020:
Please review my PR for Assignment 8 [While Loop]


harshvegada (approved):
You can handle this scenario without creating third variable, think on this.

Code match requirement but scope for improvement.

PR #129, opened 2020-10-30
RaghvendraT_assignment_8

RaghvendramaniTOct2020:


src/raghvendra/NumberOperationUsingWhile.java
```diff
@@ -0,0 +1,59 @@
+package raghvendra;
+
+public class NumberOperationUsingWhile {
+	
+	void evenNumber(int start,int end) {
+		int index=start;
```
harshvegada, line 6: You can handle this scenario without creating third variable, think on this.
can you use start++ for increment? same for all other methods

harshvegada (approved):
work on comments.

PR #130, opened 2020-10-30
AnjaliP_Asgnmt_8

AnjaliPOct2020:
Assignment 8 using While Loop

### DIFF
--- a/src/anjaliP/WhileLoop.java
+++ b/src/anjaliP/WhileLoop.java
@@ -1,0 +1,47 @@
+package anjaliP;
+
+public class WhileLoop {
+	void display1(int start,int end) {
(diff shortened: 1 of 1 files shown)


src/anjaliP/WhileLoop.java
```diff
@@ -0,0 +1,47 @@
+package anjaliP;
+
+public class WhileLoop {
+	void display1(int start,int end) {
+		System.out.println("-------------Even Numbers----------------");
+		int index=start;
```
harshvegada, line 6: You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
work on comments.
Good for merge!

PR #132, opened 2020-10-30
TejashreeT_Assignment8

TejashreeTOct2020:
Please review Assignment 8

### DIFF
--- a/src/tejashree/Assignment8While.java
+++ b/src/tejashree/Assignment8While.java
@@ -1,0 +1,63 @@
+package tejashree;
+
+public class Assignment8While {
+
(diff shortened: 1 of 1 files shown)


src/tejashree/Assignment8While.java
```diff
@@ -0,0 +1,63 @@
+package tejashree;
+
+public class Assignment8While {
+
+	void printEven(int start, int end) {
+		int index = start;
```
harshvegada, line 6: You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
work on comments.
Good for merge!

PR #133, opened 2020-10-30
BrijeshG_Assignment_8

BrijeshGOct2020:


src/brijesh/LoopAssignment8.java
```diff
@@ -0,0 +1,56 @@
+package brijesh;
+
+public class LoopAssignment8 {
+
+	void displayEvenNumber(int start, int end) {
+		System.out.println("Even numbers are:");
+		int index= start;
```
harshvegada, line 7: You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
work on comments.
Good for merge!

src/brijesh/LoopAssignment8.java
BrijeshGOct2020: Yes you are right. I can directly write something like this so we can eliminate one extra variable. Thanks for pointing out.
```
while(start<=end) {
			if(index%5==0)
				System.out.println(start);
			start++;
		}
```

PR #135, opened 2020-10-30
PriyankaK_Assignment_8

PriyankaKOct2020:
Assignment 8 : while Loop Practice

### DIFF
--- a/src/priyanka/WhileLoopExample.java
+++ b/src/priyanka/WhileLoopExample.java
@@ -1,0 +1,63 @@
+package priyanka;
+
+public class WhileLoopExample {
+
+	void evenNumber(int start, int end) {
+
+		System.out.println("**************** Even numbers are ***********");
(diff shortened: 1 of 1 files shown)


src/priyanka/WhileLoopExample.java
```diff
@@ -0,0 +1,63 @@
+package priyanka;
+
+public class WhileLoopExample {
+
+	void evenNumber(int start, int end) {
+
+		System.out.println("**************** Even numbers are ***********");
+		int index = start;
```
harshvegada, line 8: why do we need third variable, you can handle this with start variable as well, think on this.
same for all methods

harshvegada (approved):
Code looks good to me! Good for merge as matches reequipment.

Refer this code:
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/142/files

PR #137, opened 2020-10-30
NitinJ_Assignment_8

NitinJOct2020:
Write the same program using the while loop including all operations.

### DIFF
--- a/src/nitin/WhileLoop.java
+++ b/src/nitin/WhileLoop.java
@@ -1,0 +1,55 @@
+package nitin;
+
+public class WhileLoop {
+	
+	void evenNumber(int start,int end) {
+		int index = start;
+		System.out.println("Even numbers are : ");
+		while(index<=end) {
+			if(index%2==0) {
+				System.out.println(index);
+			}
+			index++;
+		}
+	}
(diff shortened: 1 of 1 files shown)


src/nitin/WhileLoop.java
```diff
@@ -0,0 +1,55 @@
+package nitin;
+
+public class WhileLoop {
+	
+	void evenNumber(int start,int end) {
+		int index = start;
```
harshvegada, line 6: why do we need extra variable here?
you can use start directly as increment variable.

harshvegada (approved):
Code matches requirement but there some scope of improvement.

Refer this code:
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/142/files

src/nitin/WhileLoop.java
NitinJOct2020: Thank you Harsh for your comment, I will try it based on your suggestion, do i have to submit the assignment again in the same pull request?

NitinJOct2020: Thank you i have tried again and it worked, i am pushing it as a practice assignment.

PR #145, opened 2020-10-30
Apurva_Asgmt_8

ApurvaQ:
Please review my PR

### DIFF
--- a/src/apurva/NumPrint.java
+++ b/src/apurva/NumPrint.java
@@ -1,0 +1,50 @@
+package apurva;
+
+class NumPrint {
+	void eveNum(int start, int end) {
+		int index = start;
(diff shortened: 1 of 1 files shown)


src/apurva/NumPrint.java
```diff
@@ -0,0 +1,50 @@
+package apurva;
+
+class NumPrint {
+	void eveNum(int start, int end) {
+		int index = start;
+		while (index <= end) {
```
harshvegada, line 6: here you need to define one extra variable, you can handle this 
if(start<=end){
//code
}
start++;

harshvegada (approved):
think on given comment, rest code looks good to me
take of this in next PR.
Good for merging.

PR #147, opened 2020-10-30
AnkitaM_Asgmnt_8

AnkitaDOct2020:
--Added Assignment8(WhileLoopExamples.java) file.

### DIFF
--- a/src/ankita/WhileLoopExamples.java
+++ b/src/ankita/WhileLoopExamples.java
@@ -1,0 +1,48 @@
+package ankita;
+
+public class WhileLoopExamples {
+	void prinEvenNumber(int start,int end) {
+		int i=start;
+		System.out.println("Even Numbers between "+ start +"&"+end);
+		while(i<=end) {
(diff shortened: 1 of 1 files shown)


src/ankita/WhileLoopExamples.java
```diff
@@ -0,0 +1,48 @@
+package ankita;
+
+public class WhileLoopExamples {
+	void prinEvenNumber(int start,int end) {
+		int i=start;
+		System.out.println("Even Numbers between "+ start +"&"+end);
+		while(i<=end) {
```
harshvegada, line 7: in while you can directly give start and will increase value of start, no need to create extra variable. think on this.

harshvegada (approved):
Code looks good to me!
Good for merge!

src/ankita/WhileLoopExamples.java
AnkitaDOct2020: Yes. Got it.

PR #163, opened 2020-10-30
PriyaM_Assignment_1

PriyaMOct2020:


src/priya/Calculator.java
```diff
@@ -0,0 +1,37 @@
+package priya;
+
+public class Calculator {
+
+	int num1 = 15;
+	int num2 = 3;
```
harshvegada, line 6: you already passing parameter in methods, so don't need to create instance variable for num1 & num2.

harshvegada (approved):
Good job @PriyaMOct2020 
Keep doing well.

PR #164, opened 2020-10-30
RubyK_Assignment_8

rubykoct2020:
RubyK_Assignment_8

### DIFF
--- a/src/ruby/ConditionWhile.java
+++ b/src/ruby/ConditionWhile.java
@@ -1,0 +1,53 @@
+package ruby;
+
+public class ConditionWhile {
+	void displayEvenNumber(int start, int end) {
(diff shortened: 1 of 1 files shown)


src/ruby/ConditionWhile.java
```diff
@@ -0,0 +1,53 @@
+package ruby;
+
+public class ConditionWhile {
+	void displayEvenNumber(int start, int end) {
+		System.out.println("Even Number between " + start + " and " + end);
+		int index = start;
```
harshvegada, line 6: You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
work on comments.
Good for merge!

PR #166, opened 2020-10-30
Shruti_Assignment_8

ShrutiSOct2020:
Assignment 8 please review and merge

### DIFF
--- a/src/shrutiS/MathWithWhile.java
+++ b/src/shrutiS/MathWithWhile.java
@@ -1,0 +1,58 @@
+package shrutiS;
+
+public class MathWithWhile {
+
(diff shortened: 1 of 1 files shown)


src/shrutiS/MathWithWhile.java
```diff
@@ -0,0 +1,58 @@
+package shrutiS;
+
+public class MathWithWhile {
+
+	void evenNumbers(int startNumber, int endNumber) {
+		System.out.println("Even Numbers are : ");
+		int evenNumber = startNumber;
```
harshvegada, line 7: You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
work on comments.
Good for merge!

PR #167, opened 2020-10-30
Anurag Assignment 8

anuragkoct2020:
This PR contains the following -

Added "WhileLoop.java" for "Assignment 8" [Ready for review]

Also added all the other files from older assignments.[Already reviewed] 

        src/anurag/Car.java
        src/anurag/ITProfessional.java
        src/anurag/RectangleCircle.java
        src/anurag/SimpleInterest.java
        src/anurag/SquareCube.java
        src/anurag/Temp.java
        src/anurag/bankAccount.java

### DIFF
--- a/src/anurag/WhileLoop.java
+++ b/src/anurag/WhileLoop.java
@@ -1,0 +1,71 @@
+package anurag;
+
+public class WhileLoop {
+
+	void even(int start, int end) {
+		int i = start;
+		System.out.print("Even numbers from range " + start + " -> " + end + " are : ");
+		while (i <= end) {
+
+			if (i % 2 == 0) {
+				System.out.print(" " + i);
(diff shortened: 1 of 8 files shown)


src/anurag/WhileLoop.java
```diff
@@ -0,0 +1,71 @@
...
+				System.out.print(" " + i);
+			}
+			i++;
+		}
+
+		System.out.println();
+	}
+
+	void divisibleBy5And3(int start, int end) {
+		int i = start;
```
harshvegada, line 33: You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
why there are 8 file changes, it should be as per the assignment program count.

work on comments.
Good for merge!

PR #169, opened 2020-10-30
KajolC_Assignment_8

KajolCOct2020:
While Loop Example

### DIFF
--- a/src/kajol/WhileLoopExample.java
+++ b/src/kajol/WhileLoopExample.java
@@ -1,0 +1,63 @@
+package kajol;
+
+public class WhileLoopExample {
+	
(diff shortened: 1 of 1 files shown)


src/kajol/WhileLoopExample.java
```diff
@@ -0,0 +1,63 @@
+package kajol;
+
+public class WhileLoopExample {
+	
+	void printEvenNumbers(int start, int end) {
+		System.out.println("Even Numbers between "+start+" and "+end+" :");
+		int index=start;
```
harshvegada, line 7: You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
work on comments.
Good for merge!

PR #170, opened 2020-10-30
RohitR_Assignment_8

RohitROct2020:


src/rohit/WhileLoop.java
```diff
@@ -0,0 +1,55 @@
+package rohit;
+
+public class WhileLoop {
+	void evennumber(int start, int end) {
+		System.out.println("Even Numbers::");
+		int index = start;
```
harshvegada, line 6: You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
work on comments.
Good for merge!

PR #171, opened 2020-10-30
Suresh_Assignment_8

sureshchukka:
Hi Harsh/Krishna,

Please Review my Assignment8 for While Loop.

Thanks & Regards,
Suresh Chukka.

### DIFF
--- a/src/suresh/While.java
+++ b/src/suresh/While.java
@@ -1,0 +1,75 @@
+package suresh;
+
+public class While {
+
+	void Even(int start, int end) {
+		System.out.println( "Even Numbers are:" );
+		int Index =start;
+		while (Index<=end) {
+			
+			if (Index % 2 == 0) {
+				
+				System.out.println(Index);
+				Index++;
+		}
+		}
+		
+	}
+	
+	void Divisible(int start, int end) {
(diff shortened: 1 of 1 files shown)


src/suresh/While.java
```diff
@@ -0,0 +1,75 @@
+package suresh;
+
+public class While {
+
+	void Even(int start, int end) {
+		System.out.println( "Even Numbers are:" );
+		int Index =start;
```
harshvegada, line 7: You can handle this scenario without creating third variable, think on this.

src/suresh/While.java
```diff
@@ -0,0 +1,75 @@
...
+		while (Index<=end) {
+			
+			if (Index % 7 ==0  ) {
+				
+				System.out.println(Index +" is Divisible by 7");
+				Index++;
+				
+		}}
+		int Index1 =start;
```
harshvegada, line 56: why do we need different while loops

Please refer: 
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/brijesh/LoopAssignment8.java

harshvegada (approved):
work on comments.
Good for merge!

PR #178, opened 2020-10-30
PoojaP_Assignment_6

PoojaPOct2020:
Please review my PR

### DIFF
--- a/src/poojap/Shop.java
+++ b/src/poojap/Shop.java
@@ -1,0 +1,121 @@
+package poojap;
+
+public class Shop {
+	
+	int qMaggie;
+	int qDosa ;
+	int qPouches; 
+	int qPanipuri;  
+	int qMasala;
+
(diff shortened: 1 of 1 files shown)


src/poojap/Shop.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+		if(qMasala >=orderMasala)
+		  qMasala = qMasala - orderMasala;
+	  	  System.out.println("Masala:Running Out of stock");
+
+	}
+	void outOfStock() {
+
+        boolean flag = false;
```
harshvegada, line 53: you dont need to use boolean value & you write multiple if condition and for all those if conditions you have one common boolean falg which is wrong logically.

Refer below code: https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/kajol/ShopStockDetails.java

harshvegada (approved):
Merging for now, pls correct understanding
let me know if you have any confusion.

PR #184, opened 2020-10-30
ShrutiC_Assignment_8

ShrutiCOct2020:
--> Code for printing even num, num divisible by 5, 3, 7 and 13 using while loop

### DIFF
--- a/src/shrutiC/MathematicalOperation.java
+++ b/src/shrutiC/MathematicalOperation.java
@@ -1,0 +1,63 @@
+package shrutiC;
+
+class MathematicalOperation {
+
+	void printEvenNos(int startIndex, int endIndex) {
+
+		System.out.println("\nEven numbers between " + startIndex + " and " + endIndex + " are :-");
+		int i = startIndex;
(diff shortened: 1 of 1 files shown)


src/shrutiC/MathematicalOperation.java
```diff
@@ -0,0 +1,63 @@
+package shrutiC;
+
+class MathematicalOperation {
+
+	void printEvenNos(int startIndex, int endIndex) {
+
+		System.out.println("\nEven numbers between " + startIndex + " and " + endIndex + " are :-");
+		int i = startIndex;
+		while (i <= endIndex) {
+			System.out.println(i);
```
harshvegada, line 10: why we need to i++ twice.
if i give range from 11 then your code will not work.

& You can handle this scenario without creating third variable, think on this.

harshvegada (approved):
work on comments.
Good for merge!

Refer this code:
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/brijesh/LoopAssignment8.java

PR #210, opened 2020-10-31
RohitR_Assignment_9

RohitROct2020:


src/rohit/BankingSystem.java
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	void creditamount(int credit) {
+		currentbalance = currentbalance + credit;
+		creditamount++;
+		allcredit++;
+	}
+
+	void printbalance() {
+
```
harshvegada, line 29: There is one requirement related this as well for individual user, that you missed here.
i want individual printTotalBalanace count as well.

harshvegada (approved):
One change request, rest logic is as expected. Keep it up!

PR #212, opened 2020-10-31
NandiniS_Assignment_9

NandiniSOct2020:


src/nandini/BankingSystem.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class BankingSystem {
+	
+	int debitCount;
+	int creditCount;
+	int currentBalance;
+	static int debitCounttotal;
+	static int creditCounttotal;
+	static int totalbalanceCount;
+	static int printTotalbalanceCount;
```
harshvegada, line 11: There is one requirement related this as well for individual user, that you missed here.
i want individual printTotalBalanace count as well.

harshvegada (approved):
One change request, rest logic is as expected. Keep it up!

PR #224, opened 2020-11-01
PrasadC_Assignment_9

PrasadCOct2020:
Hi Both,

Please review my PR.

Thanks!

### DIFF
--- a/src/prasad/Bankingsystem.java
+++ b/src/prasad/Bankingsystem.java
@@ -2,36 +2,69 @@ package prasad;
 
 public class Bankingsystem {
 	
-	long accountBalance;
-	long debitAmount;
-	long creditAmount;
-
-	void debitamount(long amount) {
-		debitAmount=amount;
-		accountBalance = accountBalance+amount;	
-		}
-
-	void creditamount(long amount) {
-		creditAmount=amount;
-		accountBalance= accountBalance-amount;
+	int accountbalance ;
+	int debitamount;
+	int creditamount;
+	int debitcount;
+	int creditcount;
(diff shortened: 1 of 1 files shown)


src/prasad/Bankingsystem.java
```diff
@@ -0,0 +1,70 @@
...
+	
+	void creditamt(int amt) {
+		creditamount= creditamount+ amt;
+		accountbalance = accountbalance+amt;
+		creditcount++;
+		totalcreditcount++;
+	}
+	
+	void debitamt(int amt) {
+		debitamount=debitamount+amt;
```
MaulikTechnocredits, line 23: why this operation ?

src/prasad/Bankingsystem.java
```diff
@@ -0,0 +1,70 @@
...
+		accountbalance = accountbalance-amt;
```
MaulikTechnocredits, line 24: what if accountbalance is less than amt ?
you should add if condition.

src/prasad/Bankingsystem.java
PrasadCOct2020: I have added this operation to calculate amount being debited from the operation so that I can display that amount in SOP. Similar operation is included to record credited amount.

src/prasad/Bankingsystem.java
PrasadCOct2020: Condition added. Pushing updated code.

Thanks for your comments!

harshvegada (approved):
Well Handle Conditions @PrasadCOct2020 
Keep it up!

PR #225, opened 2020-11-01
Anjali_Assignment9

AnjaliSoct2020:


src/anjaliS/BankingSystem.java
```diff
@@ -0,0 +1,68 @@
...
+	  static int totalDebitCnt;
+	  static int totalBalanceCnt;
+
+	   void creditAmount(int creditAmt){
+	      currentBalance=currentBalance+creditAmt;
+		  creditAmtCnt++;
+		  totalCreditCnt++;
+      }
+	    void debitAmount(int debitAmt){  
+		  currentBalance=currentBalance-debitAmt;
```
MaulikTechnocredits, line 20: If condition is required. 
what if currentBalanace is less than debitAmt ?

AnjaliSoct2020: Yes got it, will make the changes..

harshvegada (approved):
Good Job @AnjaliSoct2020 
Now it's match the requirement!

PR #244, opened 2020-11-03
SurbhiJ_Assignment_8_9

SurbhiJOct20:


src/surbhi/Assignment9Banking.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	void creditAmount(int creditAmount) {
+		currentBalance = currentBalance + creditAmount;
+		creditCnt++;
+		totalCreditCnt++;
+	}
+
+	void debitAmount(int debitAmountt) {
+		currentBalance = currentBalance - debitAmountt;
```
MaulikTechnocredits, line 22: here, you should add condition,
if (debitAmount < currentBalance)
then only deduct amount else print insufficient balance.

MaulikTechnocredits (approved):
Please take a look at comments, rest code looks good to me. Good job.

PR #274, opened 2020-11-03
NitinJ_Assignment_10

NitinJOct2020:
Created a new Package and in that created 4 classes with only 1 class have main method and other class's method are called by creating objects, combination of static and non-static function calling in between classes, also checked the Getter method using contextual help.


KrishnaKTechnocredits (approved):
Good job Nitin !!

PR #280, opened 2020-11-05
Himali Assignment11 submitted

HimaliIOct2020:


src/himali/Assignment11.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class Assignment11 {
+
+	
+	double sum(double x,double y,double z) {
+		return x+y+z;
+		
+	}
+	double average(double x,double y,double z) {
+		Assignment11 assignement11=new Assignment11();
```
harshvegada, line 11: Remove this line, as we are calling from non-static to non-static

harshvegada (changes requested):
Please do changes, video sent to you

PR #281, opened 2020-11-05
TejashreeT_Assignment11

TejashreeTOct2020:
Please review assignment11(returnType)

### DIFF
--- a/src/tejashree/Assignment11.java
+++ b/src/tejashree/Assignment11.java
@@ -1,0 +1,49 @@
+package tejashree;
+
+public class Assignment11 {
+	double sum(double x, double y, double z) {
+		return x + y + z;
+	}
+	double avg(double x, double y, double z) {
+		Assignment11 as11 = new Assignment11();
+		return as11.sum(x, y, z) / 3;
+
+	}
+	static boolean isAnsAboveExpectation(double avg) {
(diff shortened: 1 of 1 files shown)


src/tejashree/Assignment11.java
```diff
@@ -0,0 +1,49 @@
+package tejashree;
+
+public class Assignment11 {
+	double sum(double x, double y, double z) {
+		return x + y + z;
+	}
+	double avg(double x, double y, double z) {
+		Assignment11 as11 = new Assignment11();
```
harshvegada, line 8: remove object creation, as we are calling non-static method to non-static

src/tejashree/Assignment11.java
```diff
@@ -0,0 +1,49 @@
...
+
+	}
+	static boolean isAnsAboveExpectation(double avg) {
+		if (avg > 50)
+			return true;
+		else
+			return false;
+	}
+	static boolean isEligible(double avg) {
+		if (avg > 50 || avg % 2 == 0)
```
harshvegada, line 19: this should be divisible not modualar, as per the requirement.

harshvegada (changes requested):
Please accommodate required changes, video sent for same.

PR #282, opened 2020-11-05
Mrunal_Assignment11

MrunalYOct2020:


src/mrunal/Return.java
```diff
@@ -0,0 +1,52 @@
...
+  
+   
+   double sum(double x , double y, double z )
+    {
+	   return x+y+z;
+	}
+   
+   double average(double x ,double y, double z)
+   {
+	   Return return1= new Return();
```
harshvegada, line 13: Remove this line, as we are calling non-static to non-static method

harshvegada (changes requested):
Video sent for require changes

PR #284, opened 2020-11-05
PriyankaK_Assignment_11

PriyankaKOct2020:
Assignment 11 (Return Type)

### DIFF
--- a/src/priyanka/StudentGrade.java
+++ b/src/priyanka/StudentGrade.java
@@ -1,0 +1,55 @@
+package priyanka;
+
+public class StudentGrade {
+
+	double sum(int x, int y, int z) {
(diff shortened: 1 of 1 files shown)


src/priyanka/StudentGrade.java
```diff
@@ -0,0 +1,55 @@
+package priyanka;
+
+public class StudentGrade {
+
+	double sum(int x, int y, int z) {
+		return x + y + z;
+	}
+
+	double average(int x, int y, int z) {
+		StudentGrade studentGrade = new StudentGrade();
```
harshvegada, line 10: Remove this line as we are calling non-static method to non-static method

harshvegada (approved):

PR #299, opened 2020-11-05
ShrutiS_Assignment_11

ShrutiSOct2020:
image

### DIFF
--- a/src/shrutiS/StudentScore.java
+++ b/src/shrutiS/StudentScore.java
@@ -1,0 +1,65 @@
+package shrutiS;
+
+public class StudentScore {
+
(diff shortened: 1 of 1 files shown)


src/shrutiS/StudentScore.java
```diff
@@ -0,0 +1,65 @@
...
+		if (average > 50)
+			return true;
+		else
+			return false;
+	}
+
+	static boolean isEligible(double average) {
+		if (average > 50 || average % 2 == 0)
+			return true;
+		else
```
harshvegada, line 33: You can remove else as keyword part

harshvegada (approved):

PR #305, opened 2020-11-05
RitikaG_Assignment_11

RitikaGOct2020:


src/ritika/SumAndAverage.java
```diff
@@ -0,0 +1,59 @@
+package ritika;
+
+public class SumAndAverage {
+
+	double CalAverage(double num1, double num2, double num3) {
+		SumAndAverage sumAndAverage2 = new SumAndAverage();
```
harshvegada, line 6: Remove this line, as we are caling non-static to non-static method only

harshvegada (approved):

PR #316, opened 2020-11-07
AnkitJ_Assignment_12

AnkitJOct2020:
@harshvegada /@Krishna: Please review my PR for Assignment 12.

### DIFF
--- a/src/ankit/arrayConcepts/SimpleCalculation.java
+++ b/src/ankit/arrayConcepts/SimpleCalculation.java
@@ -1,0 +1,70 @@
+package ankit;
+
+public class SimpleCalculation {
+	int index = 0;
+
+	int minimumNumber(int[] arr) {
+		int minimum = arr[0];
+		for (index = 0; index < arr.length; index++) {
+			if (minimum > arr[index])
+				minimum = arr[index];
+		}
+		System.out.println("From the list of array provided by client, minimum number is : " +minimum);
+		return minimum;
+	}
+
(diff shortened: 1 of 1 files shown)


src/ankit/arrayConcepts/SimpleCalculation.java
```diff
@@ -0,0 +1,70 @@
+package ankit;
+
+public class SimpleCalculation {
+	int index = 0;
+
+	int minimumNumber(int[] arr) {
+		int minimum = arr[0];
+		for (index = 0; index < arr.length; index++) {
+			if (minimum > arr[index])
```
harshvegada, line 9: Video sent on skype for this code, pls go through it

src/ankit/arrayConcepts/SimpleCalculation.java
```diff
@@ -0,0 +1,70 @@
...
+
+	void displayReverse(int[] arr) {
+		for(index=(arr.length-1); index>=0; index--) {
+			System.out.print(arr[index]+ " ");
+		}
+		System.out.println();
+	}
+	
+	void averageOfMinAndMaxNumbers(int min, int max) {
+		int averageMinMax =0;
```
harshvegada, line 54: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.


harshvegada (approved):

PR #317, opened 2020-11-07
MilindC_Assignment_12

MIlindChavhanOct2020:
MilindC_Assignment_12 complete please review my PR

image

### DIFF
--- a/src/milind/Array.java
+++ b/src/milind/Array.java
@@ -1,0 +1,76 @@
+/*Assignment - 12 : 7th Oct'2020 
+write a single program having different methods.
+ 
+a) To find Min number from given array
+b) To find Max number from given array
+c) To find Avg of all numbers from give array
+d) To display all numbers [0 to length]
+e) To display all numbers [length to 0]
+g) Method to find average of min and max number from array.*/
+
+package milind;
+
+public class Array {
+	
+	int minNumberFromArray(int[] array) {
+		int minNumber = array[0];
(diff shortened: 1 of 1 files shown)


src/milind/Array.java
```diff
@@ -0,0 +1,78 @@
...
+e) To display all numbers [length to 0]
+g) Method to find average of min and max number from array.*/
+
+package milind;
+
+public class Array {
+	
+	int minNumberFromArray(int[] array) {
+		int minNumber = array[0];
+		for(int index = 0; index < array.length; index++) {
```
harshvegada, line 17: Video sent on skype for this code, pls go through it

Also check this code for reference 

src/milind/Array.java
```diff
@@ -0,0 +1,78 @@
...
+	
+	void printReverseArray(int[] array) {
+		System.out.print("Reverse array printing : ");
+		for(int index = array.length-1; index >= 0 ; index--) {
+			System.out.print(array[index] + ", ");
+		}
+		System.out.println("");
+	}
+	
+	double averageOfMinAndMax(int minNumberFromArray, int maxNumberFromArray) {
```
harshvegada, line 60: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.


harshvegada (approved):

PR #319, opened 2020-11-07
NandiniS_Assignment_12

NandiniSOct2020:
Added a single program to perform different Operations on Array, Please Review.

### DIFF
--- a/src/nandini/ArrayOperations.java
+++ b/src/nandini/ArrayOperations.java
@@ -1,0 +1,89 @@
+package nandini;
+
+public class ArrayOperations {
+	
+	  int getMaxNumber(int num[]) {
+		
+		int max = num[0];
+		for(int index=0; index < num.length; index++) {
+			
+			if(max < num[index])
+				max = num[index];
+		}
+		
+		return max;
+	 }
+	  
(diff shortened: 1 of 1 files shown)


src/nandini/ArrayOperations.java
```diff
@@ -0,0 +1,89 @@
+package nandini;
+
+public class ArrayOperations {
+	
+	  int getMaxNumber(int num[]) {
+		
+		int max = num[0];
```
harshvegada, line 7: Video sent on skype for this code, pls go through it

src/nandini/ArrayOperations.java
```diff
@@ -0,0 +1,89 @@
...
+				
+		   for(int index= num4.length-1; index >= 0; index--) {
+					
+					System.out.println(num4[index]);
+				}
+				     
+			}
+			
+        double getMinMaxAvg(int maxNum, int minNum) {
+
```
harshvegada, line 60: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.


harshvegada (approved):

PR #321, opened 2020-11-07
Suvela_Assignment_12

SuvelaKoct2020:
Asgmt 12

### DIFF
--- a/src/suvela/PracticeArray.java
+++ b/src/suvela/PracticeArray.java
@@ -1,0 +1,60 @@
+package suvela;
+
+public class PracticeArray {
+	int min(int[] array) {
+		int index=0;
+		int min = array[index];
+		for (index = 0; index < array.length; index++) {
+			if (array[index] < min)
+				min = array[index];
+		}
+		return min;
+
+	}
+
+	int max(int[] array) {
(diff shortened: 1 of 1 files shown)


src/suvela/PracticeArray.java
```diff
@@ -0,0 +1,60 @@
+package suvela;
+
+public class PracticeArray {
+	int min(int[] array) {
+		int index=0;
+		int min = array[index];
+		for (index = 0; index < array.length; index++) {
```
harshvegada, line 7: video sent for this code on skype, pls check & here you can write code like below as well why you define index separate n then using into for loop.

for(int index = 0;index < arr.length; index++)

src/suvela/PracticeArray.java
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	int average(int[] array) {
+		int sum = 0;
+		for (int index = 0; index < array.length; index++) {
+			sum = sum + array[index];
+		}
+		return sum / array.length;
+	}
+
```
harshvegada, line 44: looks like you missed one scenarios
g. Method to find average of min and max number from array.

PR #322, opened 2020-11-07
RubyK_Assignment_12

rubykoct2020:
image

### DIFF
--- a/src/ruby/ArrayConcepts.java
+++ b/src/ruby/ArrayConcepts.java
@@ -1,0 +1,78 @@
+/*write a single program having different methods.
+ 
+a) To find Min number from given array
+b) To find Max number from given array
+c) To find Avg of all numbers from give array
+d) To display all numbers [0 to length]
+e) To display all numbers [length to 0]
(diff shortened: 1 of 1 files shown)


src/ruby/ArrayConcepts.java
```diff
@@ -0,0 +1,78 @@
...
+		arr.averageOfMinMax(max, min);
+		
+		System.out.println("Display Array List from 0 to "+num.length);
+		arr.displayArraylist(num);
+		
+		System.out.println("Display Array List in Reverse order from "+num.length+ "to 0");
+		arr.displayReverseArraylist(num);
+	}
+	
+	void averageOfMinMax(int num1, int num2){
```
harshvegada, line 33: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.


int avgOfMinMax = (findMax(input)+findMin(input))/2

harshvegada (approved):
Very well @rubykoct2020 
this is something i am expecting.
Keep it up!

PR #323, opened 2020-11-07
Assignment 12 submitted_Himali

HimaliIOct2020:


src/himali/Assignment12.java
```diff
@@ -0,0 +1,85 @@
+package himali;
+
+public class Assignment12 {
+	int i;
+	int minimumNumber(int[] num) {
+		int min=num[0];
```
harshvegada, line 6: Video sent for this code change

src/himali/Assignment12.java
```diff
@@ -0,0 +1,85 @@
...
+		int sum = 0;
+		for(i = 0; i < num.length; i++) {
+			sum = sum + num[i];
+		}
+		double avg = sum / num.length;
+		return avg;
+		
+	}
+	int averageOfMinMaxNumber(int min,int max) {
+		int averageMinMax=0;
```
harshvegada, line 37: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.


int avgOfMinMax = (minNumber(input)+maxNumber(input))/2

harshvegada (approved):

PR #326, opened 2020-11-07
RohitR_Assignment_12

RohitROct2020:


src/rohit/Arraybasics.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	void displayallnumberslengthto0(int[] arr5) {
+		System.out.println("Array listing form length to 0- ");
+		for (int i = arr5.length - 1; i >= 0; i--) {
+			System.out.println(arr5[i]);
+		}
+	}
+
+	int averageofminmaxnumbers(int max, int min) {
```
harshvegada, line 48: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.

int avgOfMinMax = (minimumnumber(input)+maxnumber(input))/2

harshvegada (approved):

PR #327, opened 2020-11-07
Mrunal_Assignment_12

MrunalYOct2020:
image

### DIFF
--- a/src/mrunal/maxNumber.java
+++ b/src/mrunal/maxNumber.java
@@ -1,0 +1,98 @@
+/*QUESTION: write a single program having different methods.
+ 
+a) To find Min number from given array
+b) To find Max number from given array
+c) To find Avg of all numbers from give array
+d) To display all numbers [0 to length]
+e) To display all numbers [length to 0]
+g) Method to find average of min and max number from array.*/
+
+package mrunal;
+
+public class maxNumber {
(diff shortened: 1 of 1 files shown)


src/mrunal/maxNumber.java
```diff
@@ -0,0 +1,98 @@
...
+	{
+		for(index=(input.length-1) ;index>=0 ;index--)
+		{
+			System.out.println(input[index]);
+		}
+	}
+	
+	void averageOfMinMax(int[] input)
+	{
+		maxNumber m = new maxNumber();
```
harshvegada, line 64: No need to create object n then calling method, we can directly non-static method from non-static method.

src/mrunal/maxNumber.java
```diff
@@ -0,0 +1,98 @@
...
+		
+		int ans1 = m.getMinNumber(input);
+		int ans = m.getMaxNumber(input);
+		int averageOfMinMax = (ans1 + ans)/2;
```
harshvegada, line 68: You can write code like this, I remove variable and directly doing sum of min & max, no problem, if you are not comfortable to write code like below, just optimize.

harshvegada (approved):
pls go thru video comments. for understanding.

PR #328, opened 2020-11-07
ShrutiS_Assignment_12

ShrutiSOct2020:
image

### DIFF
--- a/src/shrutiS/MathWithArray.java
+++ b/src/shrutiS/MathWithArray.java
@@ -1,3 +1,14 @@
+/*Assignment - 12 : 7th Oct'2020 
+write a single program having different methods.
+ 
+a) To find Min number from given array
+b) To find Max number from given array
+c) To find Avg of all numbers from give array
+d) To display all numbers [0 to length]
+e) To display all numbers [length to 0]
+g) Method to find average of min and max number from array.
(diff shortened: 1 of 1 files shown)


src/shrutiS/MathWithArray.java
```diff
@@ -0,0 +1,84 @@
...
+
+package shrutiS;
+
+public class MathWithArray {
+	
+	int minNumber(int[] num) {
+		
+		int minNumber = num[0];
+		
+		for (int index = 0 ; index<num.length ; index++ ) {
```
harshvegada, line 20: index should starts with 1 as we are giving minNumber = num[0]

src/shrutiS/MathWithArray.java
```diff
@@ -0,0 +1,84 @@
...
+	}
+	int avgOfArrayNumber(int[] num) {
+		
+		int sum =0;
+	    int totalNumber=0;
+	    for(int index = 0; index<num.length ; index++) {
+	    	sum += num[index];
+	    	totalNumber++;
+	    }
+	    return sum/totalNumber;
```
harshvegada, line 44: you can take num.length (array's property)

harshvegada (approved):

PR #318, opened 2020-11-07
BrijeshG_Assignment_12

BrijeshGOct2020:
image

PR #330, opened 2020-11-07
RaghvendraT_assignment_12

RaghvendramaniTOct2020:
image

### DIFF
--- a/src/raghvendra/ArrayMethod.java
+++ b/src/raghvendra/ArrayMethod.java
@@ -1,0 +1,64 @@
+package raghvendra;
+
+public class ArrayMethod {
+	int min(int[] number) {
+		int min=number[0];
+		for(int index=0;index<number.length;index++) {
+			if (min>number[index])
+				min=number[index];
+		}
+		return min;
+	}
+	
+	int max(int[] number) {
+		int max=number[0];
+		for(int index=0;index<number.length;index++) {
+			if (max<number[index])
+				max=number[index];
+		}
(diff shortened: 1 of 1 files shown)


src/raghvendra/ArrayMethod.java
```diff
@@ -0,0 +1,64 @@
+package raghvendra;
+
+public class ArrayMethod {
+	int min(int[] number) {
```
harshvegada, line 4: Video sent on skype for this code, pls go through it.

src/raghvendra/ArrayMethod.java
```diff
@@ -0,0 +1,64 @@
...
+	//Found new line character from google and used it
+	void displayArrayFromEnd(int[] number) {
+		System.out.println( "\r\n" +"Elements of number Array starting from last index: ");
+		for(int index=number.length-1;index>=0;index--) {
+			System.out.print(number[index]+",");
+		}
+	}
+	
+	double avgMinMax(int min, int max) {
+		double average=0;
```
harshvegada, line 45: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.


int avgOfMinMax = (max(input)+min(input))/2

harshvegada (approved):
Logically code looks correct just some correction needed in code, which i sent you on skype.
Keep improving.

PR #331, opened 2020-11-07
RitikaG_Assignment_12

RitikaGOct2020:
image

### DIFF
--- a/src/ritika/ArrayOperations.java
+++ b/src/ritika/ArrayOperations.java
@@ -1,0 +1,81 @@
+/*
+Assignment - 12 : 7th Oct'2020
+write a single program having different methods.
+
+a) To find Min number from given array
+b) To find Max number from given array
+c) To find Avg of all numbers from give array
+d) To display all numbers [0 to length]
+e) To display all numbers [length to 0]
+g) Method to find average of min and max number from array.
+*/
+
+package ritika;
+
+public class ArrayOperations {
+
+	int findMaxNumber(int num1[]) {
+		int max = num1[0];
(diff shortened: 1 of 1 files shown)


src/ritika/ArrayOperations.java
```diff
@@ -0,0 +1,81 @@
...
+e) To display all numbers [length to 0]
+g) Method to find average of min and max number from array.
+*/
+
+package ritika;
+
+public class ArrayOperations {
+
+	int findMaxNumber(int num1[]) {
+		int max = num1[0];
```
harshvegada, line 18: Video sent on skype for this code, pls go through it

src/ritika/ArrayOperations.java
```diff
@@ -0,0 +1,81 @@
...
+		int min = num2[0];
+		for (int i = 0; i < num2.length; i++) {
+			if (num2[i] < min)
+				min = num2[i];
+		}
+		System.out.println("Min Numeber is " + min);
+		return min;
+	}
+
+	int findAvgOfMinAndMaxNumber(int min, int max) {
```
harshvegada, line 37: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.


int avgOfMinMax = (findMaxNumber(input)+findMinNumber(input))/2

harshvegada (approved):
Very well @RitikaGOct2020 
logically looks correct,
Keep it up!
Good to merge.!

PR #333, opened 2020-11-07
NagendraT_Assignment_12

NagendraTOct2020:


src/nagendra/JavaArray.java
```diff
@@ -0,0 +1,84 @@
+package nagendra;
+
+public class JavaArray {
+
+	int getMinimumNumber(int[] num) {
+
+		int minimum = num[0];
+
+		for (int index = 0; index < num.length; index++) {
```
harshvegada, line 9: Video sent on skype for this code, pls go through it

src/nagendra/JavaArray.java
```diff
@@ -0,0 +1,84 @@
...
+
+		System.out.println(" ---The Numbers from Array.length till 0 are --- ");
+
+		for (int index = num.length - 1; index >= 0; index--) {
+			System.out.println(" --->" + num[index]);
+		}
+	}
+
+	void getAverageOfMinAndMax(int minimum, int maximum) {
+		int average = (minimum + maximum) / 2;
```
harshvegada, line 59: I think we have requirement where you need to find max & min number then you need to find avg of both, so this expecting two variable in method looks wrong as per the implementation perspective, user will give us array, not min & max value. I hope you got my point.


harshvegada (approved):
Logically code looks good to me @NagendraTOct2020 
Keep it up!

PR #335, opened 2020-11-07
ShrutiC_Assignment_12

ShrutiCOct2020:
--> Array Operations

Assignment 12 Output

### DIFF
--- a/src/shrutiC/NumericOperations.java
+++ b/src/shrutiC/NumericOperations.java
@@ -1,0 +1,81 @@
+package shrutiC;
+
+public class NumericOperations {
+
+	void displayArray(int[] numberArray) {
+		System.out.print("Array : ");
+		for (int index = 0; index < numberArray.length; index++)
+			System.out.print(numberArray[index] + " ");
+	}
+
+	void displayReverseArray(int[] numberArray) {
+		System.out.print("\nReverse Array : ");
+		for (int index = numberArray.length - 1; index >= 0; index--)
+			System.out.print(numberArray[index] + " ");
+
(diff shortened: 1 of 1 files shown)


src/shrutiC/NumericOperations.java
```diff
@@ -0,0 +1,81 @@
...
+		System.out.print("\nReverse Array : ");
+		for (int index = numberArray.length - 1; index >= 0; index--)
+			System.out.print(numberArray[index] + " ");
+
+	}
+
+	int findMaximumNum(int[] numberArray) {
+
+		int maxNum = numberArray[0];
+		for (int index = 0; index < numberArray.length; index++) {
```
harshvegada, line 21: Video sent on skype for this code, pls go through it

src/shrutiC/NumericOperations.java
```diff
@@ -0,0 +1,81 @@
...
+
+	}
+
+	void findAverageOfMinMaxNum(int[] numberArray) {
+		int avg = 0;
+		int minNum, maxNum;
+		maxNum = findMaximumNum(numberArray);
+		minNum = findMinimumNum(numberArray);
+		avg = (maxNum + minNum) / 2;
+		System.out.println("Average of min and max number in an array is : " + avg);
```
harshvegada, line 62: I think we have requirement where you need to find max & min number then you need to find avg of both, user will give us array, and you need to find min & max value. and then avg of both, I hope you got my point.


harshvegada (approved):
Code looks correct, matches output!
Good to merge!

PR #339, opened 2020-11-08
AnkitaM_Exam_1_30min

AnkitaDOct2020:
--Added test1 program.
image

### DIFF
--- a/src/ankita/FindMaximumWord.java
+++ b/src/ankita/FindMaximumWord.java
@@ -18,22 +18,16 @@ package ankita;
 
 public class FindMaximumWord {
 	String findMaxWord(String[] arr) {
-		int max =0;
 		String maxWord=arr[0];
-		for(int index=0;index<arr.length;index++) {
-			String name = arr[index];
-			int cnt=0;
-			for(int i =0;i<name.length();i++) {
-				cnt = cnt +1;
-			}
-			if(max<cnt) {
-				max=cnt;
+		for(int index=1;index<arr.length;index++) {
+			if(maxWord.length()<arr[index].length()) {
 				maxWord=arr[index];
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/ankita/FindMaximumWord.java
```diff
@@ -0,0 +1,49 @@
...
+package ankita;
+
+public class FindMaximumWord {
+	String findMaxWord(String[] arr) {
+		int max =0;
+		String maxWord=arr[0];
+		for(int index=0;index<arr.length;index++) {
+			String name = arr[index];
+			int cnt=0;
+			for(int i =0;i<name.length();i++) {
```
MaulikTechnocredits, line 26: why do we need loop ????
can we directly say cnt= name.length() ???? even we dont need cnt.
dont create unnecessary vairables

src/ankita/FindMaximumWord.java
```diff
@@ -0,0 +1,49 @@
...
+				cnt = cnt +1;
+			}
+			if(max<cnt) {
```
MaulikTechnocredits, line 29: what if we remove max variable and write it like below.
if(maxWord.length()<arr[index].length())

MaulikTechnocredits (changes requested):
Please work on comments and push it again.

src/ankita/FindMaximumWord.java
AnkitaDOct2020: Code corrected.

src/ankita/FindMaximumWord.java
AnkitaDOct2020: Corrected.

MaulikTechnocredits (approved):
Excellent job Ankita. Now everything looks as expected.

PR #344, opened 2020-11-08
RubyK_Exam_1_30mins

rubykoct2020:
image

### DIFF
--- a/src/ruby/ExamMaxLength.java
+++ b/src/ruby/ExamMaxLength.java
@@ -1,0 +1,42 @@
+package ruby;
+
+import java.util.Scanner;
+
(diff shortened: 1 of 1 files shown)


MaulikTechnocredits (approved):
just a suggestion. I saw instead of storing maxName, you are storing index and from that you are again retrieving name from that index. you can directly store name instead index as we dont have any further requirements on index. 

Code looks good to me Ruby. Keep it up.

rubykoct2020:
Sure Maulik, I got it. I will keep improving myself. Thanks a lot for your encouragement. I still cant believe that I am able to write programs like this and all the credit goes to you. you are a Superb Mentor

PR #346, opened 2020-11-08
RaghvendraT_Exam1

RaghvendramaniTOct2020:
image

### DIFF
--- a/src/raghvendra/Exam1.java
+++ b/src/raghvendra/Exam1.java
@@ -1,0 +1,38 @@
+package raghvendra;
+
+import java.util.Scanner;
+
+public class Exam1 {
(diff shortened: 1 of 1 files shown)


src/raghvendra/Exam1.java
```diff
@@ -0,0 +1,39 @@
+package raghvendra;
+
+import java.util.Scanner;
+
+public class Exam1 {
+	String maxLength(String[] arr) {
+		String firstName=null;
+		firstName= arr[0];
```
MaulikTechnocredits, line 8: this can be in the same line.
String firstnName = arr[0]

MaulikTechnocredits (approved):
Code looks good to me.

PR #350, opened 2020-11-08
NandiniS_Exam_1 and the time taken is 55 minutes.

NandiniSOct2020:


src/nandini/Exam1.java
```diff
@@ -0,0 +1,51 @@
...
+	 
+	
+	public static void main(String [] args) {
+		
+		Exam1 exam1 = new Exam1();
+		String [] name = {"Nandini","Abhimanyu","Leena","Suman","Angel"};
+		String longestName = exam1.getMaxLengthName(name);
+		System.out.println("Longest Name is : "+ longestName);
+		exam1.display(longestName);
+
```
MaulikTechnocredits, line 47: I can see too many extra blank lines. please take care of it next time. line number 31,35,36, 41,28,10,15

src/nandini/Exam1.java
```diff
@@ -0,0 +1,51 @@
...
+		for(int index = 1; index < arr.length; index++) {
+			
+			if(arr[index].length() > maxLengthName) {
+				i = index;
+				maxLengthName = arr[index].length();
+			}
+			    
+		}
+
+		length = arr[i].length();
```
MaulikTechnocredits, line 18: separate length variable is not really required. we can directly use arr[i].length() in sop

src/nandini/Exam1.java
```diff
@@ -0,0 +1,51 @@
...
+		
+		return arr[i];
+	}
+	 
+	 void display(String name) {
+		 
+		 String evenCharcters = "";
+		 for(int index = 1; index < name.length(); index++) {
+			  
+			if(index % 2 == 0)
```
MaulikTechnocredits, line 29: look at the requirement. we want to print even character , index starts from 0. Odd index represents even characters of the string. look at the expected output. 

MaulikTechnocredits (approved):

PR #351, opened 2020-11-08
PoojaP_exam1_(50min)

PoojaPOct2020:
Please review my PR

### DIFF
--- a/src/poojap/CodingTest.java
+++ b/src/poojap/CodingTest.java
@@ -1,0 +1,46 @@
+package poojap;
+
+public class CodingTest {
+	
+	
+	
+	String getMaximumLength(String[] arr) {
+		{
+			String max = arr[0];
+
+			for (int index = 0; index < arr.length; index++) {
+				if (max.length() < arr[index].length()) {
+					max = arr[index];
+					
+				}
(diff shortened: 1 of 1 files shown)


src/poojap/CodingTest.java
```diff
@@ -0,0 +1,46 @@
+package poojap;
+
+public class CodingTest {
+	
+	
+
```
MaulikTechnocredits, line 6: lots of blank lines. line number 4,5,6,10,14,16,19,24,41,43.

src/poojap/CodingTest.java
```diff
@@ -0,0 +1,46 @@
...
+			}
+			return max;
+
+		}
+	}
+
+	void display(String name) {
+
+		for (int index = 0; index < name.length(); index++) {
+			if (index % 2 == 0) {
```
MaulikTechnocredits, line 26: look at the requirement. we want to print even character , index starts from 0. Odd index represents even characters of the string. look at the expected output. 

MaulikTechnocredits (approved):
Pooja, please take a look at the comments.
Rest logic looks good to me.

PR #352, opened 2020-11-08
YogitaP_Exam_1(50 min)

YogitaPOct2020:
image

### DIFF
--- a/src/yogita/Exam1.java
+++ b/src/yogita/Exam1.java
@@ -1,0 +1,27 @@
+package yogita;
+class Exam1{
+		
+   static String maximumLength(String[] arr) { 
+	   String  max = arr[0];
+	   for(int index=0;index<arr.length;index++)
+	      if(max.length()<arr[index].length())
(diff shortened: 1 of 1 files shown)


src/yogita/Exam1.java
```diff
@@ -0,0 +1,27 @@
...
+	   String  max = arr[0];
+	   for(int index=0;index<arr.length;index++)
+	      if(max.length()<arr[index].length())
+	 	    max = arr[index];
+	return max;
+	}
+	
+   void even(String name) {
+	   for(int index=0;index<name.length();index++)
+		  if(index % 2 == 0)
```
MaulikTechnocredits, line 14: look at the requirement. we want to print even character , index starts from 0. Odd index represents even characters of the string. look at the expected output. 

input : Technocredits
output : ehordt

MaulikTechnocredits (approved):
Please take a look at the comments.
Code looks good to me.

PR #354, opened 2020-11-08
VishakhaB_Assignment_13

VishakhaBOct2020:
Please review my PR for assignment 13.
image

### DIFF
--- a/src/vishakha/BankOperation.java
+++ b/src/vishakha/BankOperation.java
@@ -1,0 +1,153 @@
+package vishakha;
+
+import java.util.Scanner;
+
+public class BankOperation {
+
+	int iciciBal = 8000;
+	int hdfcBal = 5000;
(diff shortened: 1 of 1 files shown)


src/vishakha/BankOperation.java
```diff
@@ -0,0 +1,153 @@
...
+	int totalBalance;
+	static int creditTotalCount;
+	static int debitTotalCount;
+	static String bankName;
+	static String operation;
+	static int amount;
+	
+	void bankOperation(String bankName, String op, int amount){
+		switch(op){
+		case "credit":
```
harshvegada, line 28: instead of writing like this you can use op.toUpperCase(), not sure we learned or not.

harshvegada (approved):
Very well understanding, code matches requirement!
Keep it up! @VishakhaBOct2020

PR #355, opened 2020-11-08
RohitR_Exam_1

RohitROct2020:


src/rohit/Exam1.java
```diff
@@ -0,0 +1,48 @@
...
+			if (args[index].length() > maxlength.length()) {
+				maxlength = args[index];
+			}
+		}
+		return maxlength;
+	}
+
+	void displayevenchar(String maxstr) {
+		for (int index = 0; index < maxstr.length(); index++) {
+			if (index % 2 == 0) {
```
MaulikTechnocredits, line 25: look at the requirement. we want to print even character , index starts from 0. Odd index represents even characters of the string. look at the expected output. 

input : Technocredits
output : ehordt

MaulikTechnocredits (approved):
Code looks good to me. Keep it up Rohit.

PR #356, opened 2020-11-08
Rohan_Exam_01_25mins

RohanMOct2020:
Given comments are handled.
1. Loop iteration started from index -1 instead of 0.
2. Two extra spaces removed

Output is intact.

image

### DIFF
--- a/src/rohan/StringScan.java
+++ b/src/rohan/StringScan.java
@@ -32,11 +32,10 @@ public class StringScan {
 	
 	void displayEvenChar(String str) {
 		System.out.println("Display Characters at odd index: ");
-		for(int i=0;i<str.length();i++) {
+		for(int i=1;i<str.length();i++) {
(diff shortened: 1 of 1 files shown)


src/rohan/StringScan.java
```diff
@@ -32,11 +32,10 @@ String getMaxLength(String[] arr) {
 	
 	void displayEvenChar(String str) {
 		System.out.println("Display Characters at odd index: ");
-		for(int i=0;i<str.length();i++) {
+		for(int i=1;i<str.length();i++) {
 			if(!(i%2==0))
```
MaulikTechnocredits, line 36: look at the requirement. we want to print even character , index starts from 0. Odd index represents even characters of the string. look at the expected output. 

input : Technocredits
output : ehordt

MaulikTechnocredits (approved):
Code looks good to me.

PR #362, opened 2020-11-08
RitikaG_Exam_1_50mins

RitikaGOct2020:
image

### DIFF
--- a/src/ritika/Exam1.java
+++ b/src/ritika/Exam1.java
@@ -1,0 +1,51 @@
+/*
+ Coding Exam 1: 8th Nov'2020
+1) print name having maximum length and print its length also.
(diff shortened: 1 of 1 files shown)


src/ritika/Exam1.java
```diff
@@ -0,0 +1,51 @@
...
+
+}
+ */
+package ritika;
+
+public class Exam1 {
+
+	String getMaxLengthName(String[] arr) {
+		String nameMax = null;
+		int lengthMax = 0;
```
harshvegada, line 24: Here we should make lengthMax = arr[0].length();
n start loop from 1

harshvegada (approved):
Rest logic is correct
Good to merge!

PR #365, opened 2020-11-09
MonikaS_aasignment_12

MonikaSOct2020:
image
 
Assignment 12  write a single program having different methods. (Array)

### DIFF
--- a/src/monika/Array.java
+++ b/src/monika/Array.java
@@ -1,0 +1,96 @@
+
+//write a single program having different methods.
+ 
+//a) To find Min number from given array
+//b) To find Max number from given array
+//c) To find Avg of all numbers from give array
+//d) To display all numbers [0 to length]
+//e) To display all numbers [length to 0]
+//g) Method to find average of min and max number from array.
+
+package monika;
+public class Array {
+
+	  int getMaxNumber(int num[]) {
+
+		int max = num[0];
+		for(int index=0; index < num.length; index++) {
+
+			if(max < num[index])
+				max = num[index];
+		}
+
+		return max;
+	 }
+
+	  int getMinNumber(int num1[]) {
+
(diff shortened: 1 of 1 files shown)


src/monika/Array.java
```diff
@@ -0,0 +1,96 @@
...
+//e) To display all numbers [length to 0]
+//g) Method to find average of min and max number from array.
+
+package monika;
+public class Array {
+
+	  int getMaxNumber(int num[]) {
+
+		int max = num[0];
+		for(int index=0; index < num.length; index++) {
```
harshvegada, line 17: Video sent on skype for this code, pls go through it

src/monika/Array.java
```diff
@@ -0,0 +1,96 @@
...
+
+			   avg = avg + num2[index]/5;
+
+		   }
+				return avg;
+		}
+
+	  void getLengthofArray(int num3[]) {
+
+		   for(int index=0; index < num3.length; index++) {
```
harshvegada, line 51: not getting use of this method here

src/monika/Array.java
```diff
@@ -0,0 +1,96 @@
...
+		   for(int index= num4.length-1; index >= 0; index--) {
+
+					System.out.println(num4[index]);
+				}
+
+			}
+
+        double getMinMaxAvg(int maxNum, int minNum) {
+
+            double avgOfMinMax = (maxNum + minNum)/2;
```
harshvegada, line 70: I think we have requirement where you need to find max & min number then you need to find avg of both, user will give us array, and you need to find min & max value. and then avg of both, I hope you got my point.


int avgOfMinMax = (min(input)+max(input))/2

harshvegada (approved):
Please do not give much space in between two line

PR #366, opened 2020-11-09
PrasadC_Assignment_13

PrasadCOct2020:
Hi Both,

Please review my PR.
image

image

Thanks
Prasad Chitale

### DIFF
--- a/src/prasad/BankTransactions.java
+++ b/src/prasad/BankTransactions.java
@@ -1,0 +1,199 @@
+package prasad;
+
+import java.util.Scanner;
+
+public class BankTransactions {
+	float icicibalance;
+	int icicicreditcount;
+	int icicidebitcount;
+	float hdfcbalance;
+	int hdfccreditcount;
+	int hdfcdebitcount;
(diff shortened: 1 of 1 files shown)


src/prasad/BankTransactions.java
```diff
@@ -0,0 +1,199 @@
...
+			}else 
+				System.out.println("Transaction failed: Insufficient balance.");
+			break;
+		default:
+			System.out.println("Transaction failed: Invalid bank name.");
+		}
+	}
+	float totalbalance() {
+		float totalbalance=0;
+		totalbalance = icicibalance+sbibalance+hdfcbalance+bobbalance;
```
harshvegada, line 87: You can write this in single line right
float totalbalance = icicibalance+sbibalance+hdfcbalance+bobbalance;

harshvegada (approved):
Very Well @PrasadCOct2020 
I guess you are trying to some chance to user, that's why you have written 4 time transaction

just high level you need to use while loop here to give chance to end user

PR #368, opened 2020-11-09
Anurag Assignment 12

anuragkoct2020:
Assignment 12 is ready for review.

### DIFF
--- a/src/anurag/Array.java
+++ b/src/anurag/Array.java
@@ -1,0 +1,98 @@
+package anurag;
+
+public class Array {
+
+	int ar[] = { 10, 5, 30, 20, 90 };
+
+	int displayMin() {
+
+		int min = ar[0];
+
+		for (int index = 0; index < 5; index++) {
+
+			if (ar[index] < min)
+				min = ar[index];
+		}
+
+		return min;
+
+	}
+
+	int displayMax() {
+
+		int max = ar[0];
+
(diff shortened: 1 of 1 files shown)


src/anurag/Array.java
```diff
@@ -0,0 +1,98 @@
+package anurag;
+
+public class Array {
+
+	int ar[] = { 10, 5, 30, 20, 90 };
```
harshvegada, line 5: wrong thing, we need to pass array in method,
Always create parametrized method

src/anurag/Array.java
```diff
@@ -0,0 +1,98 @@
...
+
+	int displayMin() {
+
+		int min = ar[0];
+
+		for (int index = 0; index < 5; index++) {
```
harshvegada, line 11: Video sent on skype for this code, pls go through it

src/anurag/Array.java
```diff
@@ -0,0 +1,98 @@
...
+			System.out.print(ar[index]);
+			System.out.println(" ");
+
+		}
+
+	}
+
+	void minMaxArray() {
+		double avg;
+		Array cal = new Array();
```
harshvegada, line 74: Why do we need to create Object to call displayMin N Max?
you can directly call those methods as both are non-static

harshvegada (changes requested):
Logically correct but not up to mark
Please refer below code

PR #370, opened 2020-11-09
RohanM_Assignment_13

RohanMOct2020:
Please Review my PR 
image

### DIFF
--- a/src/rohan/SwitchBanking.java
+++ b/src/rohan/SwitchBanking.java
@@ -1,0 +1,159 @@
+package rohan;
+
+import java.util.Scanner;
+
+/*A person has total 4 accounts in ICICI, HDFC, BOB, SBI one in each.
+          a) Person can debit or credit amount from respective bank.
+              Hint : Switch case is expected here. [Please watch today's recording again, in case of confusion.]
+
+          b) A Method to show total balance [ICICI balance + HDFC balance + BOB balance + SBI Balance]
+
+          c) A Method to show total credit, debit operations 
+                 output : Total credit operation - 7 times
+                              Total debit operation - 5 times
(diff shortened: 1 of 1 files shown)


src/rohan/SwitchBanking.java
```diff
@@ -0,0 +1,159 @@
...
+		case "Debit":
+			switch(bankName) {
+			case "ICICI":
+						if(amount<iciciTotalBalance) {
+							iciciTotalBalance -= amount;
+							iDebitCnt++;
+						}
+						else
+							printMessage("ICICI", "Debit");
+						break;
```
harshvegada, line 75: Here, You can pass bankName, no need to pass "ICICI" as value, bcoz think when come to this case when case machted ICICI right. so in bankName variable we already have that value as ICICI/BOB etc. 
just correct thinking here

src/rohan/SwitchBanking.java
```diff
@@ -0,0 +1,159 @@
...
+		}while(choice.equalsIgnoreCase("Yes"));
+		System.out.println("----Total Balance (in all accounts)----");
+		switchBanking.displayTotalBalance();
+		System.out.println("----TotalCredit Debit Operations----");
+		switchBanking.totalCreditDebitOperations();
+		System.out.println("----Bankwise Credit Operations----");
+		switchBanking.bankwiseCreditOps();
+		System.out.println("----Bankwise Debit Operations----");
+		switchBanking.bankwiseDebitOps();
+	}
```
harshvegada, line 158: at end operation we should write scanner.close()

harshvegada (approved):
Overall code looks good to me!
Good to merge!

PR #374, opened 2020-11-09
ShrutiS_Assignment_13

ShrutiSOct2020:
image

### DIFF
--- a/src/shrutiS/CustomerBankInfo.java
+++ b/src/shrutiS/CustomerBankInfo.java
@@ -13,7 +13,7 @@
    e) A Method to show individual bank debit operations.
                 output : ICICI bank Debit Operation - 2
                          HDFC bank Debit Operation - 2
-*/
+ */
 package shrutiS;
 
 import java.util.Scanner;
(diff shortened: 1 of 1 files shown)


src/shrutiS/CustomerBankInfo.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Scanner;
+
+public class CustomerBankInfo {
+
+	// Array with balance of banks in the following order ICICI, HDFC, BOB, SBI
+	double[] balance = new double[4];
+	// Array with credits of 4 banks in the following order ICICI, HDFC, BOB, SBI
+	int noOfCredits[] = new int[4];
+	// Array with debits of 4 banks in the following order ICICI, HDFC, BOB, SBI
+	int noOfDebits[] = new int[4];
```
harshvegada, line 28: use of array here look we are increases complexity, if assume for there are 4 bank only but in future if it increases then to maintain this becomes difficult.

harshvegada: when we want some data then we ahve to iterate this array put some condition, right?
if you make this as separate variable then then it very easy to access.

harshvegada (approved):
Overall Code looks very good!
Keep it up! @ShrutiSOct2020

PR #376, opened 2020-11-09
VaibhavJ_Assignment_12

VaibhavJOct2020:
VaibhavJ_Assignment_12

### DIFF
--- a/src/vaibhav/MinMaxNumber.java
+++ b/src/vaibhav/MinMaxNumber.java
@@ -1,0 +1,75 @@
+package vaibhav;
+
+public class MinMaxNumber {
+	
+	int minNumber(int[] arr) {		
+		int min=arr[0];
+		for (int index = 0 ;index < arr.length; index++){			
+			if(arr[index] < min)
+				min=arr[index];
+			else continue;			
+		}		
+		return min;		
+	}
+	
+	int maxNumber(int[] arr) {
+		int max=arr[0];
+		for (int index = 0 ;index < arr.length; index++){			
+			if(arr[index] > max)
+				max=arr[index];
+			else continue;			
+		}		
+		return max;	
+		
+	}
+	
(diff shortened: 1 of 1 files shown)


VaibhavJOct2020:
image

src/vaibhav/MinMaxNumber.java
```diff
@@ -0,0 +1,75 @@
+package vaibhav;
+
+public class MinMaxNumber {
+	
+	int minNumber(int[] arr) {		
+		int min=arr[0];
+		for (int index = 0 ;index < arr.length; index++){
```
harshvegada, line 7: Video sent on skype for this code, pls go through it

src/vaibhav/MinMaxNumber.java
```diff
@@ -0,0 +1,75 @@
...
+		for (int index = 0 ;index < arr.length; index++){			
+			if(arr[index] > max)
+				max=arr[index];
+			else continue;			
+		}		
+		return max;	
+		
+	}
+	
+	int avgAllNumber(int [] arr) {
```
harshvegada, line 26: return type should be double

src/vaibhav/MinMaxNumber.java
```diff
@@ -0,0 +1,75 @@
...
+		int avg=0;
+		for (int index = 0 ;index < arr.length; index++ ){
+			sum+=arr[index];			
+		}
+		avg= sum / arr.length;
+		return avg;		
+	}
+	
+	int avgMinMaxNumber(int min,int max) {
+		int ans=0;
```
harshvegada, line 37: I think we have requirement where you need to find max & min number then you need to find avg of both, user will give us array, and you need to find min & max value. and then avg of both, I hope you got my point.

harshvegada (approved):
Over all looks good to me!
Logically everything is on place.
Good to merge!

PR #377, opened 2020-11-09
RaghvendraT_Assignment_13

RaghvendramaniTOct2020:
image

### DIFF
--- a/src/raghvendra/BankOperationUsingSwitch.java
+++ b/src/raghvendra/BankOperationUsingSwitch.java
@@ -1,0 +1,159 @@
+package raghvendra;
+
+import java.util.Scanner;
+
+public class BankOperationUsingSwitch {
+	static int creditCount;
+	static int debitCount;
+	static double hdfcAmount;
+	static int hdfcCreditCount;
+	static int hdfcDebitCount;
+	static int iciciDebitCount;
+	static int sbiDebitCount;
+	static int bobDebitCount;
+	static double iciciAmount;
+	static int iciciCreditCount;
+	static double bobAmount;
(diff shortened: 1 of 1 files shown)


src/raghvendra/BankOperationUsingSwitch.java
```diff
@@ -0,0 +1,153 @@
...
+			debitAmount=sc.nextDouble();
+		}
+	
+		bankSwitch(bankName, operation, creditAmount, debitAmount);
+		bankBalance(bankName);
+		bankCreditDebit();
+	}
+	
+
+	public void bankSwitch(String bankName,String operation,double creditAmount,double debitAmount) {
```
harshvegada, line 52: Hi @RaghvendramaniTOct2020 
Here think on real time scenario when we are going to bank for credit operation, that time bank not asking customer to do debit as well right?

In this method, it give impression like above scenario.

harshvegada (changes requested):
Look like you missed individual bank credits & debit operations!
Pls add those methods in class

src/raghvendra/BankOperationUsingSwitch.java
RaghvendramaniTOct2020: incorporated the changes suggested. Thanks

RaghvendramaniTOct2020:
> Look like you missed individual bank credits & debit operations!
> Pls add those methods in class

incorporated changes in new commit..pls review

harshvegada (approved):
Thanks for Quick Change @RaghvendramaniTOct2020 
Good to merge!!

PR #382, opened 2020-11-09
VaibhavJ_Assignment_13

VaibhavJOct2020:
image

### DIFF
--- a/src/vaibhav/PersonalBanking.java
+++ b/src/vaibhav/PersonalBanking.java
@@ -1,0 +1,178 @@
+package vaibhav;
+
+public class PersonalBanking {
+	
+	int iBal,sBal,hBal,bBal;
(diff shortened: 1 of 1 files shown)


src/vaibhav/PersonalBanking.java
```diff
@@ -0,0 +1,191 @@
...
+			break;
+		case "BOB" :
+			System.out.println("BOB Total balance is " +bBal);
+			break;	
+		default :
+			System.out.println("Enter Valid details to get your total balance");		
+		}		
+	}
+	
+	void totalCreditDebit(String bank) {
```
harshvegada, line 100: I guess Here you need to print totalCredit operation like this -> icic+bob+hdfc+sbi, same for debit card

harshvegada (approved):
Looks correct logically small changes require on sum of separate operation in method, which hope you understand from comments section.

Good to merge!

PR #383, opened 2020-11-10
Vaibhav_Exam_1_35mins

VaibhavJOct2020:
image

### DIFF
--- a/src/vaibhav/MaximumLength.java
+++ b/src/vaibhav/MaximumLength.java
@@ -1,0 +1,49 @@
+package vaibhav;
+
+import java.util.Scanner;
+
+public class MaximumLength {
+	int max;
+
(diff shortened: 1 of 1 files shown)


src/vaibhav/MaximumLength.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class MaximumLength {
+	int max;
+
+	String getMaxLengthName(String [] names) {
+		//System.out.println("length of Array is "+names.length);
+		max = names[0].length();
+		String maxName=names[0];
+		
+		for (int index = 0; index < names.length; index++){
```
harshvegada, line 13: Here we should make max = names[0].length();
n start loop from 1

harshvegada (approved):
Logic is up to the mark!

PR #384, opened 2020-11-10
TejashreeT_Assignment14

TejashreeTOct2020:
image

Please review

### DIFF
--- a/src/tejashree/NoOfCharInString.java
+++ b/src/tejashree/NoOfCharInString.java
@@ -10,16 +10,33 @@ class NoOfCharInString {
 				cnt++;
 			}
 		}
-		System.out.println("char" + "'" + ch + "'" + "is repeated " + cnt + " times");
+		if(cnt>0) 
+			System.out.println("char" + "'" + ch + "'" + "is repeated " + cnt + " times");
+		else
+			System.out.println("character is not present in a string");
 	}
 	public static void main(String[] args) {
 		NoOfCharInString noOfCharInString = new NoOfCharInString();
 		Scanner sc = new Scanner(System.in);
 		System.out.println("Enter a string");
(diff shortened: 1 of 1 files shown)


src/tejashree/NoOfCharInString.java
```diff
@@ -0,0 +1,42 @@
...
+		System.out.println("Enter character");
+		String temp=sc.next();
+		int count=1;
+		while(temp.length()!=1 && count<3) {
+			System.out.println("you have entered wrong character..please re-enter");
+			System.out.println("attemptsleft"+(3-count));
+			temp=sc.next();
+			count++;
+		}
+		if(temp.length()!=1)
```
harshvegada, line 33: This is condition look wrong, assume if i pass value as abcd in temp you were checking if length not equals to 1 then print Try sometime later.

src/tejashree/NoOfCharInString.java
```diff
@@ -0,0 +1,42 @@
...
+	public static void main(String[] args) {
+		NoOfCharInString noOfCharInString = new NoOfCharInString();
+		Scanner sc = new Scanner(System.in);
+		System.out.println("Enter a string");
+		String userInput = sc.next();
+		
+		System.out.println("Enter character");
+		String temp=sc.next();
+		int count=1;
+		while(temp.length()!=1 && count<3) {
```
harshvegada, line 27: Why we are checking length()!=1 it should be if length()==0 then user not given input and we want user to give input here. condition need to be update

harshvegada: assume user pass ABCD in tmep then still you are asking user to give input

harshvegada (changes requested):
SOP looks wrong here, rest code looks good to me!
please update code

PR #385, opened 2020-11-10
Surbhi_Assignment_11

SurbhiJOct20:
Please review Assignment 11.
image

### DIFF
--- a/src/surbhi/StaticNonStatic.java
+++ b/src/surbhi/StaticNonStatic.java
@@ -1,0 +1,83 @@
+package surbhi;
+
+/*Design to test static, non static, return type concepts.
+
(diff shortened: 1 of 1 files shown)


src/surbhi/StaticNonStatic.java
```diff
@@ -0,0 +1,83 @@
...
+		System.out.println(avg);
+
+		return avg;
+
+	}
+
+	static boolean isAnswerAboveExpecation(double avg1) {
+
+		boolean flag;
+		if (avg1 > 50)
```
harshvegada, line 40: if (avg1 > 50)
     return true;
return false;

harshvegada (approved):
Logically looks good

PR #396, opened 2020-11-10
AnkitaM_Asgmnt_14

AnkitaDOct2020:
image
image
image


harshvegada (approved):
Very good job
very well handle user based questions so far, keep it up!
at the end put scanner.close();

PR #401, opened 2020-11-10
Mrunal_Assignnment_14

MrunalYOct2020:


src/mrunal/freqOfChar.java
```diff
@@ -0,0 +1,55 @@
...
+			int count = 1;
+			
+			while(temp.length() != 1 && count < 3)
+				{
+					System.out.println("Please Enter only char : ");
+					temp = scanner.next();
+					count++;
+				}
+			
+			if(temp.length() != 1)
```
harshvegada, line 41: here it should be count not temp.length() right.
think on this

harshvegada (approved):
Good job @MrunalYOct2020 
Good to some improvement from previous comments,
Keep it up!
Good to merge!

PR #417, opened 2020-11-10
ShrutiS_Assignment_14

ShrutiSOct2020:
image

### DIFF
--- a/src/shrutiS/CharacterFrequency.java
+++ b/src/shrutiS/CharacterFrequency.java
@@ -1,0 +1,42 @@
+/*Assignment 14  10th Oct 2020
+Find the frequency of character from a given String array. 
+Note Take String and character from the user.
+ */
+
+package shrutiS;
+
+import java.util.Scanner;
+
+public class CharacterFrequency {
+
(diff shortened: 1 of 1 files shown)


src/shrutiS/CharacterFrequency.java
```diff
@@ -0,0 +1,42 @@
...
+		System.out.println("Enter the character to know the frequency");
+		String tempWord = sc.next();
+		int count = 1;
+		while (tempWord.length() != 1 && count < 3) {
+			System.out.println("You have entered wrong value. Please enter a character from the word provided. "+ (3 - count) + " attempts remaining");
+			tempWord = sc.next();
+			count++;
+		}
+		if (tempWord.length() != 1)
+			System.out.println("Try again later");
```
harshvegada, line 35: Here condition should be like below
if(count>3){
SOP("Try again later");
}
we should calculate this based on length() right

harshvegada (approved):
Good job @ShrutiSOct2020,
Good to some improvement from previous comments,
Keep it up!
Good to merge!

PR #426, opened 2020-11-11
AnkitaM_Asgmnt_15

AnkitaDOct2020:
--Added 3programs java files.

image
image
image
image
image

### DIFF
--- a/src/ankita/NumberFrequency.java
+++ b/src/ankita/NumberFrequency.java
@@ -17,7 +17,7 @@ public class NumberFrequency {
 		System.out.println("How many numbers you want to store ");
 		int num = sc.nextInt();
 		int[] number = new int[num];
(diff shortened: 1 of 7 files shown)


src/ankita/NumberFrequency.java
```diff
@@ -0,0 +1,37 @@
...
+package ankita;
+
+import java.util.Scanner;
+
+public class NumberFrequency {
+
+	public static void main(String[] args) {
+		Scanner sc = new Scanner(System.in);
+		System.out.println("How many numbers you want to store ");
+		int num = sc.nextInt();
```
harshvegada, line 18: always create separate method, never write whole logic in main method,

harshvegada (approved):
Thanks for quick change,
Overall all code looks good, just suggestion, never write code in main method only!

PR #432, opened 2020-11-11
Surbhi_BankingOperation.java

SurbhiJOct20:
Review my assignment

### DIFF
--- a/src/surbhi/BankingOperation.java
+++ b/src/surbhi/BankingOperation.java
@@ -1,0 +1,182 @@
+package surbhi;
+import java.util.Scanner;
+
+public class BankingOperation {
+
+	static int iBal, sBal, hBal, kBal, totalBal;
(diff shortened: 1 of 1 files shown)


src/surbhi/BankingOperation.java
```diff
@@ -0,0 +1,182 @@
...
+			System.out.println("Total No. of Credit Operations in KOTAK Bank account " + kCreditCnt);
+			break;
+		default:
+			System.out.println("Please enter valid details.");
+		}
+	}
+
+	void totalDebitCreditOperaions(String bankName) {
+		switch (bankName) {
+		case "ICICI":
```
harshvegada, line 116: here expetation was you will print like below, not in switch case. 
HDFC bank Credit Operation - 2
SBI bank Credit Operation - 1
BOB bank Credit Operation - 4

harshvegada (approved):
Very nice @SurbhiJOct20 
Code looks good to me just one comments for individual part. we not require switch case ther.

PR #434, opened 2020-11-11
Suvela_assignment_15

SuvelaKoct2020:
assign15
asgn15

### DIFF
--- a/src/suvela/NumberArray.java
+++ b/src/suvela/NumberArray.java
@@ -1,0 +1,33 @@
+package suvela;
+
+import java.util.Scanner;
+
+public class NumberArray {
+	int count = 0;
+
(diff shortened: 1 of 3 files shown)


src/suvela/NumberArray.java
```diff
@@ -0,0 +1,33 @@
+package suvela;
+
+import java.util.Scanner;
+
+public class NumberArray {
+	int count = 0;
+
+	void findfrequency(int[] array, int num) {
+		boolean flag = false;
```
harshvegada, line 9: I don't think so we need flag as variable here?
I don't see see any usage of flag variable.

harshvegada (approved):
No logical code require, code work as per the expectation.
single extra variable need to be remove from that part.

PR #435, opened 2020-11-11
NagendraT_Assignment_13

NagendraTOct2020:
Capture

### DIFF
--- a/src/nagendra/BankingOpration.java
+++ b/src/nagendra/BankingOpration.java
@@ -1,0 +1,161 @@
+package nagendra;
+
+public class BankingOpration {
+	int iciciBalance = 1000;
+	int hdfcBalance = 1000;
+	int bobBalance = 1000;
+	int sbiBalance = 1000;
+	int iciciCredit;
+	int hdfcCredit;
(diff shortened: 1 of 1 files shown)


src/nagendra/BankingOpration.java
```diff
@@ -0,0 +1,161 @@
...
+
+	void debitOperation(String bank, int amount) {
+		switch (bank) {
+		case "ICICI":
+			if (amount > iciciBalance) {
+				System.out
+						.println("The amount " + amount + " is not available in Your "+bank+" Account total Available Balance is " + iciciBalance + " ");
+			} else
+				iciciBalance -= amount;
+			iciciDebit++;
```
harshvegada, line 75: here this iciciDebit should be in bracket with else, bcoz if u see if amount > iciciBalace case still we are increasing the iciciDebit variable which looks wrong, think, you wan to increase counter when you debit from account right

harshvegada (approved):
one small change require, thats fine, i m hope you got my point.
correct in next PR with diff commit msg.
Rest code looks good to me!

PR #438, opened 2020-11-11
NitinJ_Assignment_14

NitinJOct2020:
image

Assignment 14 - Using Scanner class, Take word & char from user and print the given char & its count of occurrence in that word.

### DIFF
--- a/src/nitin/CharFrequency.java
+++ b/src/nitin/CharFrequency.java
@@ -1,0 +1,37 @@
+package nitin;
+// Assignment 14 - Find the frequency of character from a given String array.
+import java.util.Scanner;
+public class CharFrequency {
+	
+	void findFreq(String word,char ch) {
+		int count=0;
+		for(int index=0;index<word.length();index++) {
+			if(word.charAt(index)==ch) {
+				count++;
+			}
+		}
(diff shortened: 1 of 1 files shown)


src/nitin/CharFrequency.java
```diff
@@ -0,0 +1,37 @@
...
+		String word=sc.next();
+		System.out.println("Enter Char: ");
+		String temp =sc.next();
+		int count=1;
+		while(temp.length()!=1 && count<3) {
+			System.out.println("Wrong input, Please enter a single character, Attempts Left : "+(3-count));
+			temp=sc.next();
+			count++;
+		}
+		if(temp.length()!=1)
```
harshvegada, line 28: here it should be count not temp.length() right.
think on this.
we are checking count right

harshvegada (approved):
Rest logic looks good, no logical impact on that comments, just to correct understanding.
Good to merge!

src/nitin/CharFrequency.java
NitinJOct2020: Actually, As maulik explains in lecture, thats why i do it like that. However, I agreed with you.

PR #446, opened 2020-11-12
Change for Assignment 12

RohanMOct2020:
Changes done. I could have also written a method and returned to main, however since min and max were already returned and available in main, I used these values in main itself to calculate average of min and max nos from array. Hope this is okay.

image


harshvegada (approved):
Very Good @RohanMOct2020 
Quick change

PR #447, opened 2020-11-12
Changes for Assignment 13

RohanMOct2020:
Changes made to printMessgage and bankName is passed as mentioned in comments.


harshvegada (approved):
Thanks for change @RohanMOct2020

PR #458, opened 2020-11-12
AnkitJ_Assignment_16

AnkitJOct2020:
Please review my PR for assignment16

### DIFF
--- a/src/ankit/Scanner/EachCharacterFrequency.java
+++ b/src/ankit/Scanner/EachCharacterFrequency.java
@@ -1,0 +1,31 @@
+package ankit;
+
+import java.util.Scanner;
+
+public class EachCharacterFrequency {
+	
+	void scannerOperation(String word, int length) {
+		
+		for(int index =0; index<length; index++) {
+			int counter=0;
+			char character =word.charAt(index);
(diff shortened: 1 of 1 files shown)


src/ankit/Scanner/EachCharacterFrequency.java
```diff
@@ -0,0 +1,31 @@
...
+		
+		for(int index =0; index<length; index++) {
+			int counter=0;
+			char character =word.charAt(index);
+			for(int charIndex=0; charIndex<length; charIndex++) {
+				if(character == word.charAt(charIndex))
+					counter++;
+			}
+			System.out.println(character+ " is present " +counter+ " times in String");
+		}
```
harshvegada, line 17: static void scannerOperation(String word) {

		for (int index = 0; index < word.length(); index++) {
			int counter = 0;
			char character = word.charAt(index);
			for (int charIndex = 0; charIndex < word.length(); charIndex++) {
				if (character == word.charAt(charIndex))
					counter++;
			}
			System.out.println(character + " is present " + counter + " times in String");
		}
	}

harshvegada (approved):
match with expected output, no logical change require on code,

PR #468, opened 2020-11-12
RubyK_Assignment_14

rubykoct2020:
Assignment14 and previous assignment correction are done as suggested by Harsh
image


harshvegada (approved):
Thanks for change

PR #489, opened 2020-11-13
Mrunalyadav_Coding_Exam_34min

MrunalYOct2020:
image


harshvegada (commented):
Good you write separate method but in interview you always write single method to handle this kind scenarios,

Refer this code
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/490/files

PR #494, opened 2020-11-13
Suvela_Coding_Exam2

SuvelaKoct2020:
exam2


harshvegada (approved):
This is also good, but we can accommodate this requirement in single method, no need to write separate method for each.
As per the interviewer perspective this is not good approach.

Refer this code

https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/495/files

PR #497, opened 2020-11-13
RohitR_Coding_Exam_2

RohitROct2020:


harshvegada (approved):
You can handle this all in single loop as well!
No need to write for loop for each method method,
Think on this.

Check this
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/brijesh/codingexams/Exam2.java

PR #511, opened 2020-11-13
Priya m assigmt 16

PriyaMOct2020:
image

@harshvegada please review assignment 16


harshvegada (approved):
I can see two file with same logic just class name changed,
May be duplicate.
Code looks good to me
Good to merge!

PR #521, opened 2020-11-13
Nitin j assignment 15

NitinJOct2020:
Program 1 - Find the frequency of character from a given String array.
image

Program 2 - Program to Count the Number of Vowels in user defined string.

Positive flow Execution Result -
image

Negative flow Execution Result -
image


harshvegada (commented):
very well  @NitinJOct2020 ,
Good to merge!
No comments require! Logic is met the expectations.

PR #544, opened 2020-11-16
PrasadC_Assignment_19

PrasadCOct2020:
Hi Both,
Please review my PR.

O/P SS:-
image

image

### DIFF
--- a/src/prasad/Reversestringarray.java
+++ b/src/prasad/Reversestringarray.java
@@ -1,0 +1,45 @@
+package prasad;
+
+import java.util.Scanner;
+
+public class Reversestringarray {
+	
+	void reverseprint(String[]array){
+		String[]reverse = new String[array.length];
+		for(int i=array.length-1 , j=0 ; i>=0 && j<reverse.length; i--,j++) {
+			reverse[j]=array[i];
+	}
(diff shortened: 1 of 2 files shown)


src/prasad/Reversestringarray.java
```diff
@@ -0,0 +1,45 @@
...
+		int count = scanner.nextInt();
+		String [] words = new String[count];
+		for(int index0 = 0;index0<count;index0++) {
+			System.out.println("Please enter a word:");
+			String word = scanner.next();
+			words[index0]=word;
+		}
+		System.out.println("Given string array:-");
+		for(int index=0;index<words.length;index++) {
+			if(index<words.length-1)
```
harshvegada, line 32: why do we need here if condition @PrasadCOct2020

harshvegada (approved):
Code Logic is correct, Do not add more printing statement in code, if you are debugging in your then it's file once, code working as expected then you should remove SOP statement from code.

PrasadCOct2020:
Hi Harsh,

I added that if block to print , in between words and not to print , when we are printing last word of the string array. Hence we get printing like 
Credit,Techno,from,Diwali,Happy(,) --- here I wanted to avoid printing , hence that if else block.

PR #546, opened 2020-11-16
ShrutiS_Assignment_18

ShrutiSOct2020:
Program1
image

Program2
image

### DIFF
--- a/src/shrutiS/assignment18/MaxFrequencyCharacter.java
+++ b/src/shrutiS/assignment18/MaxFrequencyCharacter.java
@@ -5,21 +5,26 @@ output : Hi -> 3*/
 package shrutiS.assignment18;
 
 public class MaxFrequencyCharacter {
-
+	int maxFrequency;
+	String word;
 	void maxFrequencyCharCount(String input) {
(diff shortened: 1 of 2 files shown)


src/shrutiS/assignment18/MaxFrequencyCharacter.java
```diff
@@ -0,0 +1,36 @@
...
+	int maxFrequency;
+	String word;
+	void maxFrequencyCharCount(String input) {
+		String[] arr = input.split(" ");
+		for (int index = 0; index < arr.length; index++) {
+		    word = arr[index];
+			int count = 0;
+			if (input.contains(word)) {
+				for (int innerIndex = 0; innerIndex < arr.length; innerIndex++) {
+					if (word.equals(arr[innerIndex])) {
```
harshvegada, line 17: Here you already apply .equals() method then no need to use contains() method, think here, you are checking twice for same word, & you not putted else condition for contains()  method condition. 

here you should remove this condition.

harshvegada (approved):
No impact on expected output, but need to improve code efficiency.
Good to merge!
Pls take of this in next PR

PR #547, opened 2020-11-16
ShrutiS_Assignment_19

ShrutiSOct2020:
Program 1:
image

Program 2:
image

### DIFF
--- a/src/shrutiS/assignment19/ReverseStringArray.java
+++ b/src/shrutiS/assignment19/ReverseStringArray.java
@@ -1,0 +1,30 @@
+/*Program : 2 
+reverse given string array.
+input : {"Credits", "Techno","From","Diwali","Happy"}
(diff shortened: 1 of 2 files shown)


src/shrutiS/assignment19/ReverseStringArray.java
```diff
@@ -0,0 +1,30 @@
...
+output : {"Happy","Diwali","From","Techno","Credits"}*/
+
+package shrutiS.assignment19;
+
+public class ReverseStringArray {
+
+	void reverseStringArray(String[] input) {
+		String[] temp = new String[input.length];
+		int count = 0;
+		System.out.print("\nReverse of given String Array : ");
```
harshvegada, line 13: as you are not return reverse of array  & you are printing array's word here only, then I think you should not create new array and filling value into that array.

harshvegada (approved):
Code Looks good to me!
Added comments for creating new array

PR #555, opened 2020-11-16
Priyanka_Assignment_18

PriyankaKOct2020:
Program 1:
image

Program 2:
image

### DIFF
--- a/src/priyanka/Assignment_18/MaxRepeatingWord.java
+++ b/src/priyanka/Assignment_18/MaxRepeatingWord.java
@@ -1,0 +1,38 @@
+package priyanka.Assignment_18;
+
+/*Program : 1
+max repeating word with its freq.
+input : Hi Hello Hi Techno Hello Hi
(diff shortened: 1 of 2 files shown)


src/priyanka/Assignment_18/MaxRepeatingWord.java
```diff
@@ -0,0 +1,38 @@
...
+
+	void maxRepeatWord(String str) {
+
+		String[] strArr = str.split(" ");
+		int max = 0;
+		String maxWord = "";
+		for (int index = 0; index < strArr.length; index++) {
+			int count = 0;
+			String word = strArr[index];
+			if (str.contains(word)) {
```
harshvegada, line 18: Here you are putting one word from array to the variable called "word", & then you are checking that variable value into array, so here think logically, it will always gives you true right?

harshvegada (approved):
Code matches with require output, but need to correct logically.
Good to merge!
take of this next time!

PR #560, opened 2020-11-16
Apurva_Asgmt_18

ApurvaQ:
Program_1: Screenshot: https://prnt.sc/vk80dj
Program_2 Screenshot: https://prnt.sc/vk80uw

### DIFF
--- a/src/apurva/Program_1.java
+++ b/src/apurva/Program_1.java
@@ -1,0 +1,33 @@
+package apurva;
+/*Max repeating word with its frequency.
+input: Hi Hello Hi Techno Hello Hi
+output: Hi -> 3*/
+
+public class Program_1 {
+	void displayCount(String str) {
+		int max = 0, maxWord = 0;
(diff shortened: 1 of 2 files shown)


src/apurva/Program_1.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class Program_1 {
+	void displayCount(String str) {
+		int max = 0, maxWord = 0;
+		String[] strArray = str.split(" ");
+		for (int i = 0; i < strArray.length; i++) {
+			String word = strArray[i];
+			int count = 0;
+			if (str.contains(word)) {
+				for (int j = 0; j < strArray.length; j++) {
```
harshvegada, line 14: Here you are putting one word from array to the variable called "word", & then you are checking that variable value into array, so here think logically, it will always gives you true right?

harshvegada (approved):
Good job ,
code matches expected output!
Keep it up!
Good to merge!
take of this next time

PR #562, opened 2020-11-16
Vaibhav j assignment 18

VaibhavJOct2020:
program -->2

image

program -->1

image

### DIFF
--- a/src/vaibhav/MaxOccuranceWord.java
+++ b/src/vaibhav/MaxOccuranceWord.java
@@ -1,0 +1,32 @@
+package vaibhav;
+
+public class MaxOccuranceWord {
+	
+	static void occuranceWord(String str) {
+		String [] splitted=str.split(" ");
+		int max = 0;
+		String maxName = null; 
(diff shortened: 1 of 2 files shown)


src/vaibhav/MaxOccuranceWord.java
```diff
@@ -0,0 +1,39 @@
...
+	
+	static void occuranceWord(String str) {
+		String [] splitted=str.split(" ");
+		int max = 0;
+		String maxName = null; 
+		for (int index = 0; index < splitted.length;index++) {
+			int count = 0;
+			String word = splitted[index];	
+			if (str.contains(word)) {
+				for ( int innerIndex = 0 ; innerIndex < splitted.length ; innerIndex++) {
```
harshvegada, line 20: Here you are putting one word from array to the variable called "word", & then you are checking that variable value into array, so here think logically, it will always gives you true right?

harshvegada (approved):
very well @VaibhavJOct2020 
Good to merge!
No comments require! Logic is met the expectations.

PR #574, opened 2020-11-16
RaghvendraT_assignment_18

RaghvendramaniTOct2020:
image
image

### DIFF
--- a/src/raghvendra/StringAssignments/MaxRepeatingWord.java
+++ b/src/raghvendra/StringAssignments/MaxRepeatingWord.java
@@ -1,0 +1,39 @@
+package raghvendra.StringAssignments;
+/*Program : 1
+max repeating word with its freq.
+input : Hi Hello Hi Techno Hello Hi
+output : Hi -> 3*/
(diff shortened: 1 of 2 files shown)


src/raghvendra/StringAssignments/MaxRepeatingWord.java
```diff
@@ -0,0 +1,39 @@
...
+public class MaxRepeatingWord {
+	void RepeatingWordFrequency(String name) {
+		String [] arr=name.split(" ");
+		int count=0;
+		int maxCount=0;
+		int maxIndex=0;
+		for(int outerIndex=0;outerIndex<arr.length;outerIndex++) {
+			String firstWord=arr[outerIndex];
+			count=0;
+			if(name.contains(firstWord)) {
```
harshvegada, line 16: Here you are putting one word from array to the variable called "word", & then you are checking that variable value into array, so here think logically, it will always gives you true right?

harshvegada (approved):
Rest logic is correct,
Good to merge!

PR #578, opened 2020-11-17
MonikaS_Assignment_18

MonikaSOct2020:


src/monika/MaxOccuranceWord.java
```diff
@@ -0,0 +1,41 @@
...
+	public class MaxOccuranceWord {
+
+	static void occuranceWord(String str) {
+		String [] splitted=str.split(" ");
+		int max = 0;
+		String maxName = null; 
+		for (int index = 0; index < splitted.length;index++) {
+			int count = 0;
+			String word = splitted[index];	
+			if (str.contains(word)) {
```
harshvegada, line 21: Here you are putting one word from array to the variable called "word", & then you are checking that variable value into array, so here think logically, it will always gives you true right?

harshvegada (approved):
Rest code looks good to me!
Just want to correct thinking on this.

PR #601, opened 2020-11-17
RaghvendraT_assignment_21

RaghvendramaniTOct2020:
image

PR #608, opened 2020-11-17
Prasad_C_Assignment_21

PrasadCOct2020:
Hi Both,

Please review my PR,

image

Regards 
Prasad Chitale

### DIFF
--- a/src/prasad/Identicalarray.java
+++ b/src/prasad/Identicalarray.java
@@ -1,0 +1,59 @@
+package prasad;
+/*WAP to find given 2 arrays are identical or not. 
+int[] arr1 = {10,12,55,32,17};
+int[] arr2 = {10,12,55,32,17}};
+output : Identical
+--------------------------------------------
+int[] arr1 = {10,12,55,32,17,99};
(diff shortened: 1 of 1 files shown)


src/prasad/Identicalarray.java
```diff
@@ -0,0 +1,59 @@
...
+output : Not Identical*/
+
+public class Identicalarray {
+	boolean scanarray(int[]arraya,int[]arrayb) {
+			boolean flag=false;
+			if(arraya.length==arrayb.length) {
+				for(int index=0;index<arraya.length;index++) {
+					if(arraya[index]==arrayb[index])
+						flag=true;
+					else {
```
harshvegada, line 22: you should put break in if condition, think here logically you want to break when arrays value is not matches. otherwise its fine & here i feel we need only if condition require.

harshvegada (changes requested):
Refer this code & acknowledge  me.
Pls do changes
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/601/files

harshvegada:
Code not updated,
Merging.

PR #610, opened 2020-11-17
RubyK_Assignment_20

rubykoct2020:
image


harshvegada (approved):
Very well work @rubykoct2020 
As per the expectations, keep it up!!

PR #611, opened 2020-11-17
Asgmt_18

DeepakLOct2020:
@harshvegada , please review and merge the same

image

image

### DIFF
--- a/src/deepak/MaxRepeatingWordWithFreq.java
+++ b/src/deepak/MaxRepeatingWordWithFreq.java
@@ -1,0 +1,36 @@
+/* max repeating word with its freq.
+input : Hi Hello Hi Techno Hello Hi
+output : Hi -> 3 */
+
+package deepak;
+
+public class MaxRepeatingWordWithFreq {
+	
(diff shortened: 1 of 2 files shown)


src/deepak/MaxRepeatingWordWithFreq.java
```diff
@@ -0,0 +1,36 @@
...
+public class MaxRepeatingWordWithFreq {
+	
+	static void freq(String str) {
+		String [] input=str.split(" ");
+		int max = 0;
+		String maxName = null; 
+		for (int index = 0; index < input.length;index++) {
+			int count = 0;
+			String word = input[index];	
+			if (str.contains(word)) {
```
harshvegada, line 16: Here you are putting one word from array to the variable called "word", & then you are checking that variable value into array, so here think logically, it will always gives you true right?

harshvegada (approved):
Both code looks good to me. just need to correct understanding
no changes require! good to merge
Keep it up! @DeepakLOct2020

PR #613, opened 2020-11-17
Asgmt_21

DeepakLOct2020:
@harshvegada , plz review and merge the same

image

### DIFF
--- a/src/deepak/ArrayIdenticalOrNot.java
+++ b/src/deepak/ArrayIdenticalOrNot.java
@@ -1,0 +1,37 @@
+package deepak;
+
+public class ArrayIdenticalOrNot {
+	static boolean flag;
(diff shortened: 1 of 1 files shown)


src/deepak/ArrayIdenticalOrNot.java
```diff
@@ -0,0 +1,37 @@
...
+			for(int index=0; index < ar1.length; index++) {
+				
+				if(ar1[index] == ar2[index])
+					flag = true;
+				else
+					flag = false;
+			}
+		} 
+		
+		if(flag == true) {
```
harshvegada, line 20: We should not check == with true, look not logical
because that flag value already have some value like true/false.

harshvegada (approved):
Very well handle code, @DeepakLOct2020 
Keep it up!

PR #630, opened 2020-11-17
Anjali_Assignment_12

AnjaliPOct2020:
Assignment 12 
topic - Array


harshvegada (approved):
I am expecting something like this 
Very Good @AnjaliSoct2020 
Keep it doing well!
Good to merge!

PR #632, opened 2020-11-17
NagendraT_Assignment_21

NagendraTOct2020:

### DIFF
--- a/src/nagendra/IdenticalArray.java
+++ b/src/nagendra/IdenticalArray.java
@@ -1,0 +1,29 @@
+package nagendra;
+
+public class IdenticalArray {
+
+	void verifyIdenticalArray(int[] array1, int[] array2) {
+		boolean flag = true;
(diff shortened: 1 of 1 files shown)


src/nagendra/IdenticalArray.java
```diff
@@ -0,0 +1,29 @@
+package nagendra;
+
+public class IdenticalArray {
+
+	void verifyIdenticalArray(int[] array1, int[] array2) {
+		boolean flag = true;
+		if (array1.length == array2.length) {
+			for (int i = 0; i < array1.length; i++) {
+				if (array1[i] == array2[i])
```
harshvegada, line 9: You can put condition like below
if (array1[i] != array2[i]){
      flag = false;
      break;
}

harshvegada (approved):
Once you get flag value false then ideally you should break loop, as the array's are not identical right.
No logical change, just highlighting.

PR #655, opened 2020-11-17
Sadhana_Assignment_18

SadhanaPOct2020:
@harshvegada @KrishnaKTechnocredits Please review my PR

### DIFF
--- a/src/sadhana/RepeatedWordCount.java
+++ b/src/sadhana/RepeatedWordCount.java
@@ -1,0 +1,39 @@
+package sadhana;
+
+/*
+Assignment_18
+max repeating word with its freq.
+input : Hi Hello Hi Techno Hello Hi
+output : Hi -> 3
+*/
+
+public class RepeatedWordCount {
+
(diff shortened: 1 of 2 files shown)


src/sadhana/RepeatedWordCount.java
```diff
@@ -0,0 +1,39 @@
...
+public class RepeatedWordCount {
+
+	void printRepeatedWordCount(String string) {
+		System.out.println("Given string: " + string);
+		int max = 0, maxWordIndex = 0;
+		String[] strArray = string.split(" ");
+		for (int index = 0; index < strArray.length; index++) {
+			String word = strArray[index];
+			int counter = 0;
+			if (string.contains(word)) {
```
harshvegada, line 19: Here, you are pick up word value from array and checks that word is in string(method param) & then in inner for you are checking .equals(), just need to correct.

harshvegada (approved):
Over all code looks good to me,
Just to correct understanding,
Good to merge!

PR #688, opened 2020-11-18
Nagendra t assignment 22

NagendraTOct2020:


harshvegada (approved):
Very well @NagendraTOct2020 
Code Looks good to me, correct logic is applied
Keep doing well,
Good to merge!

PR #697, opened 2020-11-19
Apurva_Asgmt_22

ApurvaQ:
Please Review My PR, Screenshot: https://prnt.sc/vm5edp

### DIFF
--- a/src/apurva/PrimeNumber.java
+++ b/src/apurva/PrimeNumber.java
@@ -1,0 +1,30 @@
+package apurva;
+
+import java.util.Scanner;
+
+//Print all prime numbers between 2 to 100.
+public class PrimeNumber {
+	void printPrime(int num) {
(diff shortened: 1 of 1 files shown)


src/apurva/PrimeNumber.java
```diff
@@ -0,0 +1,30 @@
+package apurva;
+
+import java.util.Scanner;
+
+//Print all prime numbers between 2 to 100.
+public class PrimeNumber {
+	void printPrime(int num) {
+		String primeNumbers = "";
+		for (int index = 2; index <= num; index++) {
```
harshvegada, line 9: requirement is user will two limit, but looks like you give only one limit which is upper.
No change require just check requirement.

harshvegada (approved):
Code looks good to me!
Good to merge!

PR #698, opened 2020-11-19
VishakhaB_Assignment_22

VishakhaBOct2020:
image

### DIFF
--- a/src/vishakha/FindPrimeNos.java
+++ b/src/vishakha/FindPrimeNos.java
@@ -1,0 +1,29 @@
+/*Print all prime numbers between 2 to 100*/
+
+package vishakha;
+
+import java.util.Scanner;
+
+public class FindPrimeNos {
+	
+	void findPrimeNos(int startIndex, int endIndex){
+		for(int i=startIndex; i<=endIndex; i++){
+			int count = 0;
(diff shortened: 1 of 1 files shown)


harshvegada (changes requested):
video sent for correction

src/vishakha/FindPrimeNos.java
```diff
@@ -0,0 +1,29 @@
...
+package vishakha;
+
+import java.util.Scanner;
+
+public class FindPrimeNos {
+	
+	void findPrimeNos(int startIndex, int endIndex){
+		for(int i=startIndex; i<=endIndex; i++){
+			int count = 0;
+			for(int innerIndex=1; innerIndex<=endIndex; innerIndex++){
```
MaulikTechnocredits, line 12: This is not correct !!
we should not start with 0 and for every number we dont want to run the loop till endIndex.
if number is 17, we should start loop from 2 and go till Math.sqrt(17).

MaulikTechnocredits: consider start to end range is from 1 to 2000.
when you process number 1000 is prime or not, you will divide it with 2 then 3 then 4 so on..
once it is divided with any number other than 1, you should break the loop (otherwise time complexity would be worst) and declare number is not prime.

PR #701, opened 2020-11-19
Surbhi j assignment 15 16

SurbhiJOct20:
Please review my Assignment 15 &16

### DIFF
--- a/src/surbhi/CountVowels.java
+++ b/src/surbhi/CountVowels.java
@@ -1,0 +1,61 @@
+package surbhi;
+
+/*Java Program to Count the Number of Vowels in user defined string.
+
+Input : technocredits
+output : vowels are e , o, i 
+*/
+
+import java.util.Scanner;
+
+public class CountVowels {
+
+	void countVowels(String string) {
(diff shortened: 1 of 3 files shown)


src/surbhi/CountVowels.java
```diff
@@ -0,0 +1,61 @@
...
+
+	void countVowels(String string) {
+		int aCnt = 0, eCnt = 0, iCnt = 0, oCnt = 0, uCnt = 0;
+		String vowels = "";
+		for (int index = 0; index < string.length(); index++) {
+			char character = string.charAt(index);
+			switch (character) {
+			case 'A':
+			case 'a':
+				if (++aCnt == 1)
```
harshvegada, line 21: What is use of ++aCnt ==1 here?
what if when same char occurred again.
remove this if condition from here, simply you need to increases the cnt value why do you think we need to put if condition here?

harshvegada (changes requested):
Please update code

harshvegada (approved):
Not not updated as per the comment,
Merging

PR #713, opened 2020-11-20
AnkitaM_Asgmnt_23

AnkitaDOct2020:
image


harshvegada (approved):
very good Ankita,
but in interview do not take array or anything, else some counter questions will start there only,
good to merge!

PR #714, opened 2020-11-20
PoojaP_Assignment_22

PoojaPOct2020:
image

### DIFF
--- a/src/poojap/PrimeNumberr.java
+++ b/src/poojap/PrimeNumberr.java
@@ -2,26 +2,26 @@ package poojap;
 
 public class PrimeNumberr {
 
-	void findPrimeNumber(int start, int end) {
+	void findPrime(int start, int end) {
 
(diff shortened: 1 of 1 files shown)


src/poojap/PrimeNumberr.java
```diff
@@ -0,0 +1,27 @@
...
+
+		System.out.println("Prime Numbers from " + start + " to " + end);
+
+		for (int i = start; i <= end; i++) {
+			int num = i;
+			boolean flag =true;
+			for (int j = 2; j <= num / 2; j++) {
+				if (num % j == 0)
+					flag = false;
+			}
```
harshvegada, line 15: here, we should break the loop, because, we know flag value update to false only when number is not prime number, then why we will iterate aging for given number.

harshvegada (approved):
code looks good me,
matches requirement
good to merge

PR #716, opened 2020-11-20
Suvela_Coding_exam3

SuvelaKoct2020:
exam3

### DIFF
--- a/src/suvela/Exam3.java
+++ b/src/suvela/Exam3.java
@@ -1,0 +1,46 @@
+package suvela;
+
+public class Exam3 {
+	int count;
+
+	void duplicateElement(int[] arr) {
+		for (int i = 0; i < arr.length; i++) {
+			count = 0;
+			int no = arr[i];
+			for (int j = 0; j < arr.length; j++) {
+				if (no == arr[j]) {
+					count++;
+					arr[i] = 0;
+				}
+
+			}
+			if (count > 1 && arr[i] == 0)
+				System.out.println("Duplicate elements are:" + no);
+		}
+	}
+
+	void duplicateString(String[] s) {
+		for (int i = 0; i < s.length; i++) {
+			count = 0;
+			String s1 = s[i];
+			for (int j = 0; j < s.length; j++) {
+				if (s1 == s[j]) {
+					count++;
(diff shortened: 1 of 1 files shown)


src/suvela/Exam3.java
```diff
@@ -0,0 +1,46 @@
...
+			count = 0;
+			int no = arr[i];
+			for (int j = 0; j < arr.length; j++) {
+				if (no == arr[j]) {
+					count++;
+					arr[i] = 0;
+				}
+
+			}
+			if (count > 1 && arr[i] == 0)
```
MaulikTechnocredits, line 17: why we need arr[i] == 0 condition ? I think only count condition would be sufficient.

src/suvela/Exam3.java
```diff
@@ -0,0 +1,46 @@
...
+				System.out.println("Duplicate elements are:" + no);
+		}
+	}
+
+	void duplicateString(String[] s) {
+		for (int i = 0; i < s.length; i++) {
+			count = 0;
+			String s1 = s[i];
+			for (int j = 0; j < s.length; j++) {
+				if (s1 == s[j]) {
```
MaulikTechnocredits, line 27: string should not be compared with == .
you should use .equals method

src/suvela/Exam3.java
```diff
@@ -0,0 +1,46 @@
...
+					count++;
+					s[i] = null;
+				}
+			}
+			if (count > 1 && s[i] == null)
```
MaulikTechnocredits, line 32: s[i] == null is not required.

MaulikTechnocredits (changes requested):
PLease change code as suggested.

src/suvela/Exam3.java
SuvelaKoct2020: ok, got it.

src/suvela/Exam3.java
SuvelaKoct2020: changed.

src/suvela/Exam3.java
SuvelaKoct2020: got it.

PR #718, opened 2020-11-20
YogitaP_Assignment_23

YogitaPOct2020:
image

### DIFF
--- a/src/yogita/PrimeNumbersOperation.java
+++ b/src/yogita/PrimeNumbersOperation.java
@@ -1,0 +1,43 @@
+/* Assignment_23 : 20th Nov'2020
+
+1) Find count of prime numbers between 2-100.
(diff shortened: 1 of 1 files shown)


src/yogita/PrimeNumbersOperation.java
```diff
@@ -0,0 +1,43 @@
...
+				if (num % j == 0) {
+					flag = true;
+					break;
+				}
+			}
+			if (!flag) {
+				System.out.print(index + " ");
+				cnt++;
+				sum += cnt;
+				avg = sum / cnt;
```
harshvegada, line 30: we can perform this operation after the all loop iteration.

harshvegada (approved):
code looks good me,
matches requirement
good to merge

PR #722, opened 2020-11-20
TejashreeT_Assignment23

TejashreeTOct2020:
image

please review

### DIFF
--- a/src/tejashree/Assignment23.java
+++ b/src/tejashree/Assignment23.java
@@ -1,0 +1,37 @@
+package tejashree;
+
+public class Assignment23 {
+
+	public static void main(String a[]) {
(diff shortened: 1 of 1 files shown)


src/tejashree/Assignment23.java
```diff
@@ -0,0 +1,37 @@
...
+				if (i % j == 0) {
+					isPrime = false;
+					break;
+				}
+			}
+			if (isPrime) {
+				cnt++;
+				System.out.print(i + " ");
+				sum = sum + i;
+				avg = sum / cnt;
```
harshvegada, line 27: this avg function we should outside of loop,

harshvegada (approved):
very well handle scenarios,
keep it up!
code looks good to me!

PR #725, opened 2020-11-20
RubyK_Assignment_23

rubykoct2020:
image


harshvegada (approved):
writing modular way always better, but keep in mind when you are infront of interviewer and if this kind scenarios they then try accommodate their requirement in single method.

very well handle scenarios,
keep it up!
code looks good to me!

PR #727, opened 2020-11-20
NitinJ_Assignment_22

NitinJOct2020:
image

### DIFF
--- a/src/nitin/PrimeNumber.java
+++ b/src/nitin/PrimeNumber.java
@@ -1,0 +1,19 @@
+package nitin;
+/* Assignment-22 : 18th Nov'2020 - Print all prime numbers between 2 to 100.*/
+public class PrimeNumber {
+	public void findPrimeNumber(int start,int end) {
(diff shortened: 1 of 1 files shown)


src/nitin/PrimeNumber.java
```diff
@@ -0,0 +1,19 @@
+package nitin;
+/* Assignment-22 : 18th Nov'2020 - Print all prime numbers between 2 to 100.*/
+public class PrimeNumber {
+	public void findPrimeNumber(int start,int end) {
+		System.out.println("Prime Numbers Between : "+start+" to "+end+" are -->");
+		for(int index=start;index<end;index++) {
+			boolean flag=true;
+			for(int inner=2;inner<Math.sqrt(index);inner++) {
+				if(index%inner==0) 
+					flag=false;
```
harshvegada, line 10: here we should put break, because you know you come in if condition only in case of if that number is not prime right, once you know that part you don't need to check with further number. just code efficiency

harshvegada (approved):
code looks good me,
matches requirement
good to merge

PR #734, opened 2020-11-20
Surbhi j assignment 17 to 21

SurbhiJOct20:
Please review my assignment.

### DIFF
--- a/src/surbhi/WordCount.java
+++ b/src/surbhi/WordCount.java
@@ -1,0 +1,32 @@
+package surbhi;
+public class WordCount {
+
+	void wordFrequencyCount(String string) {
+		System.out.println("Given string: " + string);
+		int max = 0, maxWordIndex = 0;
+		String[] strArray = string.split(" ");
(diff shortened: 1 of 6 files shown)


src/surbhi/WordCount.java
```diff
@@ -0,0 +1,32 @@
...
+public class WordCount {
+
+	void wordFrequencyCount(String string) {
+		System.out.println("Given string: " + string);
+		int max = 0, maxWordIndex = 0;
+		String[] strArray = string.split(" ");
+		for (int index = 0; index < strArray.length; index++) {
+			String word = strArray[index];
+			int counter = 0;
+			if (string.contains(word)) {
```
harshvegada, line 11: Here contains method makes no sense, as you checking "word" in "string" (from the string you get this keyword so definitely that key word is in "string"), just logically correction

harshvegada (approved):
over all code looks good me,
matches requirement
good to merge

PR #738, opened 2020-11-20
Vaibhav j assignment 23

VaibhavJOct2020:
image

### DIFF
--- a/src/vaibhav/PrimeNumberAddition.java
+++ b/src/vaibhav/PrimeNumberAddition.java
@@ -1,5 +1,7 @@
 package vaibhav;
 
+//This program will find prime array and sum and average of all elements
+
 public class PrimeNumberAddition {
 	
 	static double finalCount=0;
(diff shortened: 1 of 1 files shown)


src/vaibhav/PrimeNumberAddition.java
```diff
@@ -0,0 +1,44 @@
...
+
+//This program will find prime array and sum and average of all elements
+
+public class PrimeNumberAddition {
+	
+	static double finalCount=0;
+	
+	public static int[] primeNumber() {
+		int i,j,k=0;
+		int [] primeArray=new int[100];
```
harshvegada, line 11: If tomorrow some one gave range more then 100 that time this will not work. instead of updating in array and then you are doing sum of those array element, better to sum in if condition it self, try to avoid array during if you can handle this scenario in single method, mostly in interview.

src/vaibhav/PrimeNumberAddition.java
```diff
@@ -0,0 +1,44 @@
...
+		for (i = 2 ; i <=100 ;i++) {
+			int count =0;
+			for (j = 1 ; j<= i; j++) {
```
harshvegada, line 14: Video sent for this, pls update this

harshvegada (changes requested):
update code, and let me know

PR #739, opened 2020-11-20
AnjaliS_Assignment22

AnjaliSoct2020:
image

### DIFF
--- a/src/anjaliS/PrimeNumbersInRange.java
+++ b/src/anjaliS/PrimeNumbersInRange.java
@@ -1,0 +1,29 @@
+package anjaliS;
+
+import java.util.Scanner;
+
+public class PrimeNumbersInRange {
+	void displayPrime(int start, int end) {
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Code looks good to me,
Good to merge!
Video sent for improvement.

src/anjaliS/PrimeNumbersInRange.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Scanner;
+
+public class PrimeNumbersInRange {
+	void displayPrime(int start, int end) {
+		for (int index = start; index <= end; index++) {
+			boolean flag = false;
+			for (int j = 2; j <= index / 2; j++) {
+				if (index % j == 0)
+					flag = true;
+			}
```
harshvegada, line 12: Here you should use break, because once you know agr if condition get satisfies then that number is not prime number.
then no need to check further iterations

PR #741, opened 2020-11-20
AnjaliS_Assignment23

AnjaliSoct2020:
image

### DIFF
--- a/src/anjaliS/PrimeNumbersAvgSum.java
+++ b/src/anjaliS/PrimeNumbersAvgSum.java
@@ -1,0 +1,38 @@
+package anjaliS;
+
+import java.util.Scanner;
+
+public class PrimeNumbersAvgSum {
+	void PrimeAvgSum(int start, int end) {
+		int sum = 0, avg = 0, count = 0;
+		for (int index = start; index <= end; index++) {
+			boolean flag = false;
+			for (int j = 2; j <= index / 2; j++) {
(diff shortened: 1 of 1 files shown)


src/anjaliS/PrimeNumbersAvgSum.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Scanner;
+
+public class PrimeNumbersAvgSum {
+	void PrimeAvgSum(int start, int end) {
+		int sum = 0, avg = 0, count = 0;
+		for (int index = start; index <= end; index++) {
+			boolean flag = false;
+			for (int j = 2; j <= index / 2; j++) {
+				if (index % j == 0)
+					flag = true;
```
harshvegada, line 12: same from previous assignment comment.

src/anjaliS/PrimeNumbersAvgSum.java
```diff
@@ -0,0 +1,38 @@
...
+			}
+			if (!flag) {
+				System.out.print(index + " ");
+				count++;
+				sum += count;
+				avg = sum / count;
```
harshvegada, line 18: once loop is end then you should perform avg formula, why we need to do operation for each iteration.

harshvegada (approved):

harshvegada:
Code matches requirement & output, as discuss, take of this next time
Good to merge

PR #747, opened 2020-11-20
AnjaliP_Assignment_22

AnjaliPOct2020:
Assignment-22 : 18th Nov'2020 
Print all prime numbers between 2 to 100.
image

### DIFF
--- a/src/anjaliP/PrimeNumber.java
+++ b/src/anjaliP/PrimeNumber.java
@@ -1,0 +1,22 @@
+package anjaliP;
+/*Assignment-22 : 18th Nov'2020 
+Print all prime numbers between 2 to 100.
+ */
+public class PrimeNumber {
+	
+	public void findPrimeNumber(int start,int end) {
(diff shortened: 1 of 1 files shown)


src/anjaliP/PrimeNumber.java
```diff
@@ -0,0 +1,22 @@
...
+public class PrimeNumber {
+	
+	public void findPrimeNumber(int start,int end) {
+		System.out.println("Prime Numbers Between : "+start+" to "+end+" are -->");
+		for(int index=start;index<end;index++) {
+			boolean flag=true;
+			for(int innerIndex=2;innerIndex<Math.sqrt(index);innerIndex++) {
+				if(index%innerIndex==0) 
+					flag=false;
+			}
```
harshvegada, line 14: Here, you should use break, because once you know this number is not prime then why we need to check that number with different vlaues

harshvegada (approved):
Just one code improvement point, pls take of this next time

PR #757, opened 2020-11-22
KajolC_Assignment_23

KajolCOct2020:
Assignment-23-FindCountSumAndAverageofPrimeNumbers

### DIFF
--- a/src/kajol/FindCountSumAndAverageofPrimeNumbers.java
+++ b/src/kajol/FindCountSumAndAverageofPrimeNumbers.java
@@ -1,0 +1,40 @@
+package kajol;
+
+import java.util.Scanner;
+
+public class FindCountSumAndAverageofPrimeNumbers {
+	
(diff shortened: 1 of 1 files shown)


src/kajol/FindCountSumAndAverageofPrimeNumbers.java
```diff
@@ -0,0 +1,40 @@
...
+			boolean flag=true;
+			for(int innerIndex=2;innerIndex<=Math.sqrt(index);innerIndex++) {
+				if(index%innerIndex==0)
+					flag=false;
+			}
+			if(flag) {
+				count++;
+				System.out.print(index+" ");
+				sum=sum+index;
+				average=sum/count;
```
harshvegada, line 21: this average formula you can do after for loop iterations, at end of loop we need to check this operation. dont'need to check avg on every iteration.

harshvegada (approved):
Just correction at one point.
Rest code looks good to me!

PR #771, opened 2020-11-23
Ankita_Asgmnt_27

AnkitaDOct2020:
image
image


harshvegada (approved):
Very well Handle both scenario,
Good to merge!

PR #783, opened 2020-11-23
MonikaS_Assignment_23

MonikaSOct2020:


src/monika/PrimeCountAvgSum.java
```diff
@@ -0,0 +1,45 @@
...
+
+	void avgSumPrimeNumbers(int startRange, int endRange) {
+		int cnt=0,sum=0;
+		float avg=0f;
+		if(startRange > endRange || startRange == endRange)
+			System.out.println("Please enter exact range");
+		else {
+			for(int index=startRange;index<=endRange;index++) {
+				boolean flag = true;
+				for(int innerIndex=2;innerIndex<index;innerIndex++) {
```
harshvegada, line 20: You should iterate till Math.sqrt(index) only, we cover during class.

harshvegada (approved):
Code matches requirement & output
Code looks good to me!
Good to merge

PR #798, opened 2020-11-23
mrunal_Assignment18

MrunalYOct2020:


src/mrunal/prog1.java
```diff
@@ -0,0 +1,42 @@
...
+ static void displayCount(String str) 
+ {
+			String [] strArray=str.split(" ");
+			int max = 0;
+			String maxName = null; 
+			for (int index = 0; index < strArray.length;index++)
+			{
+				int count = 0;
+				String word = strArray[index];	
+				if (str.contains(word))
```
harshvegada, line 18: Hi Mrunal, here you pick word from array n check whether that is in main string or not, this word definitely exist in that main string right?
So here contains condition not make sense much

harshvegada (approved):
Rest code logic is correct,
code looks good to me!

PR #802, opened 2020-11-23
Mrunal_Assignment22

MrunalYOct2020:
image

### DIFF
--- a/src/mrunal/prim_2_to_100.java
+++ b/src/mrunal/prim_2_to_100.java
@@ -1,0 +1,26 @@
+package mrunal;
+/* Assignment-22 : 18th Nov'2020 
+Print all prime numbers between 2 to 100.*/
+
+public class prim_2_to_100 
+{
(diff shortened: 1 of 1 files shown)


src/mrunal/prim_2_to_100.java
```diff
@@ -0,0 +1,26 @@
...
+	void findPrimeNumber(int start,int end) 
+	{
+		System.out.println("Displaying prime numbers  : "+start+" to "+end+" are-");
+		for(int index=start;index<end;index++) 
+		{
+			boolean flag=true;
+			for(int innerIndex=2;innerIndex<Math.sqrt(index);innerIndex++) 
+			{
+				if(index%innerIndex==0) 
+					flag=false;
```
harshvegada, line 16: Once you get falg value false then, you should break the loop as you know that number not prime number anymore.

harshvegada (approved):
very well written code
logic is correct
Good to merge!

PR #810, opened 2020-11-23
MilindC_Assignment_23

MIlindChavhanOct2020:
MilindC_Assignment_23 complete please review my PR

### DIFF
--- a/src/milind/PrimeNumberAndAdditionAndAverage.java
+++ b/src/milind/PrimeNumberAndAdditionAndAverage.java
@@ -1,0 +1,34 @@
+package milind;
+
+import java.util.Arrays;
+
+public class PrimeNumberAndAdditionAndAverage {
+	
+	void primeNumber(int startRange, int endRange) {
+		int sum = 0;
+		int countPrimeNumber = 0;
+		for(int index = startRange; index <= endRange; index++) {
+			int count = 0;
+			for(int innerIndex = startRange; innerIndex <= index; innerIndex++) {
+				if(index % innerIndex == 0) {
+					count++;
+				}
+			}
+			if(count <= 1) {
+				countPrimeNumber++;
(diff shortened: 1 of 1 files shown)


src/milind/PrimeNumberAndAdditionAndAverage.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Arrays;
+
+public class PrimeNumberAndAdditionAndAverage {
+	
+	void primeNumber(int startRange, int endRange) {
+		int sum = 0;
+		int countPrimeNumber = 0;
+		for(int index = startRange; index <= endRange; index++) {
+			int count = 0;
+			for(int innerIndex = startRange; innerIndex <= index; innerIndex++) {
```
harshvegada, line 12: Here, innerIndex always starts from 2

src/milind/PrimeNumberAndAdditionAndAverage.java
```diff
@@ -0,0 +1,34 @@
...
+				if(index % innerIndex == 0) {
+					count++;
+				}
```
harshvegada, line 15: once you know this number will not be prime number then you can put break in if condition.

src/milind/PrimeNumberAndAdditionAndAverage.java
```diff
@@ -0,0 +1,34 @@
...
+			}
+			if(count <= 1) {
```
harshvegada, line 17: here you should write only < condition
after doing above code change think for 13
13%2
13%3 n so on for no single number if condition will satisfied so cnt will be 0,
better to use boolean flag instead of int counter

harshvegada (changes requested):
Rest code logic looks good to me!
Pls update code ASAP & acknowledge me.

PR #821, opened 2020-11-23
RohitR_Assignment_24

RohitROct2020:


src/rohit/Sum.java
```diff
@@ -0,0 +1,26 @@
+package rohit;
+
+/*program 1: sum of all digits from a given string
+input: te1ch9kj8
+output:= 18
+*/
+public class Sum {
+
+	void isDigit(String str) {
```
harshvegada, line 9: Never give method name which is given by Java, might confused you.

harshvegada (approved):
Code looks good to me!

PR #825, opened 2020-11-24
PrasadC_Assignment_25

PrasadCOct2020:
Hi Both,
Please review my PR.
image

Regards 
Prasad Chitale


PrasadCOct2020:
Hi @harshvegada / @MaulikTechnocredits I will commit another class which does sum of whole numbers and not digits.

PrasadCOct2020:
Hi @MaulikTechnocredits  I have added this condition else if(Character.isLetter(ch)|| !Character.isLetterOrDigit(ch)) 
for scenario when we have a number 100 which is followed by a space. In such case we will add the captured number to sum and make temp ="";

PrasadCOct2020:
Eg: "I have 15 years and **100** 2 months of experience"

harshvegada (approved):
Very well prasad,
Code looks good to me!
Good to merge!

PR #841, opened 2020-11-24
NagendraT_Assignment_25

NagendraTOct2020:

### DIFF
--- a/src/nagendra/string/SumOfNumbers.java
+++ b/src/nagendra/string/SumOfNumbers.java
@@ -1,0 +1,36 @@
+package nagendra.string;
+
+/*
+Program 1: sum of all numbers from give string.  
+String msg = I have 15 years and 2 months of experience.
+*/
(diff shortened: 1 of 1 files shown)


src/nagendra/string/SumOfNumbers.java
```diff
@@ -0,0 +1,36 @@
...
+String msg = I have 15 years and 2 months of experience.
+*/
+public class SumOfNumbers {
+
+	void getSumOfNumbers(String words) {
+		String[] word = words.split(" ");
+		boolean flag = false;
+		int totalSum = 0;
+		String input = "";
+		for (int i = 0; i < word.length; i++) {
```
harshvegada, line 14: you dont need this for loop,

harshvegada (approved):
using only single loop you can do this coding
apply consecutives number logic here. and acknowledge me on skype

PR #848, opened 2020-11-24
Surbhi j assignment 23

SurbhiJOct20:
Please review my Assignment

### DIFF
--- a/src/surbhi/Assignment23.java
+++ b/src/surbhi/Assignment23.java
@@ -43,7 +43,7 @@ public class Assignment23 {
 
 	public static void main(String[] args) {
 		Assignment23 assignment23 = new Assignment23();
-		System.out.println(assignment23.primeSum(2, 100));
+		System.out.println(assignment23.primeSum(2, 200));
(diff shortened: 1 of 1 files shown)


src/surbhi/Assignment23.java
```diff
@@ -0,0 +1,49 @@
+package surbhi;
+public class Assignment23 {
+
+	static boolean checkPrime(int numberToCheck) {
+		if (numberToCheck == 1) {
```
harshvegada, line 5: very good catch

src/surbhi/Assignment23.java
```diff
@@ -0,0 +1,49 @@
...
+			return false;
+		}
+		for (int i = 2; i * i <= numberToCheck; i++) {
+			if (numberToCheck % i == 0) {
```
harshvegada, line 9: you shoudl iterate till Math.sqrt(numberToCheck)

harshvegada (approved):
Very good coding, very good improvement
keep it up!

PR #858, opened 2020-11-24
Suvela_assignment_24

SuvelaKoct2020:
a24

### DIFF
--- a/src/suvela/Assignment24.java
+++ b/src/suvela/Assignment24.java
@@ -1,0 +1,28 @@
+/*Assignment - 24 : 22nd Nov'2020 
+
+program 1: sum of all digits from a given string  
+input: te1ch9kj8 
+output:= 18*/
+package suvela;
+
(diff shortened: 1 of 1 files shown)


harshvegada (commented):
Missing code

src/suvela/Assignment24.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class Assignment24 {
+	void getSumOfDigits(String input) {
+		int i=0;
+		int sum=0;
+		for(int index=0;index<input.length();index++) {
+			char ch=input.charAt(index);
+			boolean flag=Character.isDigit(ch);
+			if(flag) {
+				i = Character.getNumericValue(ch);
```
harshvegada, line 16: no need to define extra variable "i" as you can directly do add operation on sum

harshvegada (approved):
Code looks good to me!
matches requirement

PR #860, opened 2020-11-24
Suvela asgn25

SuvelaKoct2020:
a25

### DIFF
--- a/src/suvela/Assignment25.java
+++ b/src/suvela/Assignment25.java
@@ -1,0 +1,35 @@
+
+/*Assignment - 25 : 22nd Nov'2020 
+
+Program 1: sum of all numbers from give string.  
+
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
this code will give expected output, but time complexity is more!
you can handle this in single loop as well

void getSum(String input) {
		int sum = 0;
		String tmp = "";
		for (int j = 0; j < input.length(); j++) {
			char ch = input.charAt(j);
			if (Character.isDigit(ch)) {
				tmp += ch;
			} else if (tmp != "") {
				sum += Integer.parseInt(tmp);
				tmp = "";
			}
		}
		if (tmp != "")
			sum += Integer.parseInt(tmp);
		System.out.println("The sum of digits is :" + sum);
	}

PR #866, opened 2020-11-24
Vaibhav_j_Assignment 25

VaibhavJOct2020:
image

### DIFF
--- a/src/vaibhav/GetIntFromString.java
+++ b/src/vaibhav/GetIntFromString.java
@@ -1,5 +1,13 @@
 package vaibhav;
 
+/*Program 1: sum of all numbers from give string.
+
+String msg = I have 15 years and 2 months of experience.
+
+hint : Integer.parseInt(String input) will help to convert string to int.
+you may need spilt, charAt method as well.*/
+
+
 public class GetIntFromString {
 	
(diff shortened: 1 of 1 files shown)


src/vaibhav/GetIntFromString.java
```diff
@@ -0,0 +1,32 @@
...
+
+hint : Integer.parseInt(String input) will help to convert string to int.
+you may need spilt, charAt method as well.*/
+
+
+public class GetIntFromString {
+	
+	public static void getInt(String[] splittedString) {
+		int sum = 0;
+		for (int index = 0 ; index < splittedString.length; index++) {
```
harshvegada, line 15: You don't require this, you can handle this in single loop

void getInt(String input) {
		int sum = 0;
		String tmp = "";
		for (int j = 0; j < input.length(); j++) {
			char ch = input.charAt(j);
			if (Character.isDigit(ch)) {
				tmp += ch;
			} else if (tmp != "") {
				sum += Integer.parseInt(tmp);
				tmp = "";
			}
		}
		if (tmp != "")
			sum += Integer.parseInt(tmp);
		System.out.println("The sum of digits is :" + sum);
	}

harshvegada (approved):
No logical change require,
matches requirement.

PR #871, opened 2020-11-24
RubyK_assignment_25

rubykoct2020:
image


src/ruby/SumOfNumInStatement.java
```diff
@@ -0,0 +1,42 @@
...
+				char ch = str.charAt(i);
+				if (Character.isDigit(ch))
+					temp += ch;
+				else if (!temp.equals("")) {
+					num += Integer.parseInt(temp);
+					temp = "";
+				}
+			}
+		}
+		if (!temp.equals("")) {
```
harshvegada, line 31: Good catch for boundary value

harshvegada (approved):
Code looks good to me!
Good to merge!

PR #887, opened 2020-11-25
PoojaP_Assignment_28

PoojaPOct2020:
image

PR #888, opened 2020-11-25
BrijeshG_Assignment_28

BrijeshGOct2020:
image

### DIFF
--- a/src/brijesh/assignment28/ArrayOperation.java
+++ b/src/brijesh/assignment28/ArrayOperation.java
@@ -1,0 +1,49 @@
+package brijesh.assignment28;
+import java.util.Arrays;
+
+/*
+find the triplet having maximum sum.
+[10,12,8,20,18,22,15,13,17]
+output : maximum sum triplet 20,18,22 --> 60
+*/
+
+public class ArrayOperation {
+	void findTripletWithMaximumSum(int[] input) {
+		System.out.println("Given Array: "+Arrays.toString(input)+"\nLength of Array: "+input.length);
+		int max=0,maxIndex=0;
+		for(int index=0; index<input.length; index=index+3) {
+			int sum=0;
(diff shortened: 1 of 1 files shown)


harshvegada (changes requested):
Hi Brijesh,
This scenario can be handle in single loop as well,

check below code.
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/887/files

BrijeshGOct2020:
Hi @harshvegada,

Thanks for comment. What if my array has 11 elements. I think code #887 will not work in such case. That's why I have used 2 for loop to handle those scenarios.
Let me know your view.

harshvegada:
I agree with your point @BrijeshGOct2020 , for 12 element that will not work, 
I think we have requirement for array which has length in multiplication of 3, then only that code will work, So either we put condition like below
if(arrat.length % 3 == 0){
    //code logic for triplet sum
} else{
    SOP("Enter valid length array");
}

If there requirement to handle all type of array then i need to check again.

BrijeshGOct2020:
Hi @harshvegada ,
Please check and let me know if this logic is correct for all type of arrays.

src/brijesh/assignment28/ArrayOperation.java
```diff
@@ -0,0 +1,49 @@
...
+			if(index<input.length)
+				System.out.print(input[index]+" ");
+		System.out.println("\n******************************************************************* ");
+	}
+	
+	public static void main(String[] args) {
+		int[] input1= {10,12,8,20,18,22,15,13,17,10};
+		int[] input2= {10,12,8,20,18,22,15,50};
+		int[] input3= {0,100,20,30,80,60,7,90,56,18,12,89};
+		int[] input4= {20,10};
```
harshvegada, line 40: We cannot consider this as triplet right

harshvegada (approved):

PR #891, opened 2020-11-25
AnkitJ_Assignment_28

AnkitJOct2020:
Please review my PR for Assignment 28


harshvegada (changes requested):
Hi @AnkitJOct2020 
I checked code but looks like it not matches requirement, at end I want to print that triplet value which you missed I guess in code,
pls take look of below code. Do changes ASAP

https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/pull/887/files

harshvegada (approved):
Code logic looks good, still i feel check this

https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/poojap/TripletSum.java

PR #896, opened 2020-11-25
ShrutiS_Assignment_22

ShrutiSOct2020:
image

### DIFF
--- a/src/shrutiS/assignment22/PrimeNumbers.java
+++ b/src/shrutiS/assignment22/PrimeNumbers.java
@@ -1,0 +1,25 @@
+/*Assignment-22 : 18th Nov'2020 
+Print all prime numbers between 2 to 100.*/
+package shrutiS.assignment22;
+
+public class PrimeNumbers {
+
+	void findPrimeNumbers() {
+		for (int index = 2; index <= 100; index++) {
+			int count = 0;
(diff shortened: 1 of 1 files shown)


src/shrutiS/assignment22/PrimeNumbers.java
```diff
@@ -0,0 +1,25 @@
+/*Assignment-22 : 18th Nov'2020 
+Print all prime numbers between 2 to 100.*/
+package shrutiS.assignment22;
+
+public class PrimeNumbers {
+
+	void findPrimeNumbers() {
+		for (int index = 2; index <= 100; index++) {
+			int count = 0;
+			for (int innerIndex = 1; innerIndex <= index / 2; innerIndex++) {
```
harshvegada, line 10: Hi @ShrutiSOct2020 
We cannot consider as prime or non-prime so ideally loop should start from 2 onwards.

src/shrutiS/assignment22/PrimeNumbers.java
```diff
@@ -0,0 +1,25 @@
...
+				if (index % innerIndex == 0)
+					count++;
```
harshvegada, line 12: once you are if condition after that you should break the loop, bcoz now you knwo that number is not prime number, right

harshvegada (approved):
Pls take of this next time, 
Keep practice of coding.
Good to merge

PR #913, opened 2020-11-25
RubyK_Assignment_27

rubykoct2020:
image

### DIFF
--- a/src/ruby/NumberInString.java
+++ b/src/ruby/NumberInString.java
@@ -1,0 +1,67 @@
+package ruby;
+/*
+------------------------------------------
+Assignment - 27: 23rd Nov'2020  
+
+Program 1: print all the words which does't have any digit. 
+String[] input = {"Masgj3Pw", "r4fsjk","techno","sfjk44lk"};
+output : techno */
+
+public class NumberInString {
+
+	void findNumberInString(String[] input) {
+		System.out.println("Program 1: Output");
+		for (int index = 0; index < input.length; index++) {
+			String str = input[index];
+			boolean flag = false;
+			for (int i = 0; i < str.length(); i++) {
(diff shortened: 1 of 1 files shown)


src/ruby/NumberInString.java
```diff
@@ -0,0 +1,67 @@
...
+	void findNumberInString(String[] input) {
+		System.out.println("Program 1: Output");
+		for (int index = 0; index < input.length; index++) {
+			String str = input[index];
+			boolean flag = false;
+			for (int i = 0; i < str.length(); i++) {
+				char ch = str.charAt(i);
+				if (Character.isDigit(ch)) {
+					flag = true;
+				}
```
harshvegada, line 21: you should break once you know that string contains digit

src/ruby/NumberInString.java
```diff
@@ -0,0 +1,67 @@
...
+			String str = input[index];
+			String temp = "";
+			int sum = 0;
+			for (int i = 0; i < str.length(); i++) {
+				char ch = str.charAt(i);
+				if (Character.isDigit(ch)) {
+					temp += ch;
+				}
+				if (!temp.equals("")) {
+					sum += Integer.parseInt(temp);
```
harshvegada, line 46: you can use Character.getNumericValue(ch) method right,
you need to add single char value and do sum

harshvegada: I hope will clear on above method, take of this next
No logical change require

harshvegada (approved):
Code matches require output!
Code logically correct!
Good to merge!

src/ruby/NumberInString.java
rubykoct2020: Thanks Harsh for you comments. I will correct and push it again

PR #941, opened 2020-11-26
NandiniS_Assignment_29

NandiniSOct2020:
Added a java file to find second largest number without sorting. Please review.

### DIFF
--- a/src/nandini/SecondMaxNumber.java
+++ b/src/nandini/SecondMaxNumber.java
@@ -1,0 +1,26 @@
+package nandini;
+
+public class SecondMaxNumber {
+	
+	void findSecondMaxNum(int[] numArr) {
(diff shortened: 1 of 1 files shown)


src/nandini/SecondMaxNumber.java
```diff
@@ -0,0 +1,26 @@
+package nandini;
+
+public class SecondMaxNumber {
+	
+	void findSecondMaxNum(int[] numArr) {
+		int largest = 0;
```
harshvegada, line 6: you can put
firstMax = input[0]
secondMax = input[1]
& start loop from 2

harshvegada (approved):
Code matches require output!
Code looks logically correct!
Good to merge!

PR #942, opened 2020-11-26
AnjaliS_Assignment28

AnjaliSoct2020:
image


harshvegada (approved):
Code looks Good to me!
Code looks logical correct!

we looking for 2nd largest only, but good you try something else as well

PR #943, opened 2020-11-26
RubyK_Assignment_28

rubykoct2020:
image

### DIFF
--- a/src/ruby/Assignment28.java
+++ b/src/ruby/Assignment28.java
@@ -1,0 +1,32 @@
+package ruby;
+/*Assignment - 28 : 25th Nov'2020 
+
+find the triplet having maximum sum.
+[10,12,8,20,18,22,15,13,17]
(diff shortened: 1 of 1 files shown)


harshvegada (changes requested):
Looks like you code not working for below array
int array[] = { 10, 12, 8, 20, 18, 22, 15, 19, 20, 25 };

Please refer this code:
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/poojap/TripletSum.java

harshvegada (approved):
Code looks good to me!
Still check this code & acknowledge me!
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/priyanka/Assignment28/MaxSumOfTriplet.java

rubykoct2020:
@harshvegada the above refer program will also work only for array which have set of 3 numbers else getting error array out of bound. as in SOP we are using index+1 and index+2

PR #947, opened 2020-11-27
Suvela_coding_exam4

SuvelaKoct2020:
exam4_1
exam4_2

### DIFF
--- a/src/suvela/exams/Exam4_2.java
+++ b/src/suvela/exams/Exam4_2.java
@@ -1,0 +1,49 @@
+
+/*Program 2: WAP to find two String are identical or not
+
+input 1: 
+String s1 = "technoCredits";
+String s2 = "teChnoCreditS";
+
+output 1: both string are not identical
+
+input 2:
+String s1 = "pune";
+String s2 = "pune";
+
+output 2: both string are identical*/
+package suvela.exams;
+
+public class Exam4_2 {
+	void isStringIdentical(String s, String s2) {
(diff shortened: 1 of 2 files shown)


harshvegada (changes requested):
I don't think so you require two for loops here,
using single loop you can check,

using charAt() you can check, hope you get this hint

SuvelaKoct2020:
Please check now. @harshvegada

src/suvela/exams/Exam4_2.java
```diff
@@ -0,0 +1,49 @@
...
+
+input 2:
+String s1 = "pune";
+String s2 = "pune";
+
+output 2: both string are identical*/
+package suvela.exams;
+
+public class Exam4_2 {
+	void isStringIdentical(String s, String s2) {
```
harshvegada, line 18: Ideally you should this code
think and then write logic


void isStringIdentical(String s, String s2) {
		System.out.println("given strings are:" + s + "," + s2);
		boolean flag = false;
		if (s.length() == s2.length()) {
			for (int i = 0; i < s.length(); i++) {
				if (s.charAt(i) == s2.charAt(i))
					flag = true;
				else {
					flag = false;
					break;
				}
			}
		}
		if (flag)
			System.out.println("both strings are identical");
		else
			System.out.println("both strings are not identical");
	}

harshvegada (approved):
more practice require on coding

PR #950, opened 2020-11-27
Tejashree coding exam 4

TejashreeTOct2020:
image

image

image

### DIFF
--- a/src/tejashree/CompareTwoArrays.java
+++ b/src/tejashree/CompareTwoArrays.java
@@ -13,13 +13,12 @@ package tejashree;
 public class CompareTwoArrays {
 
 	public void compareArrays(int []a1, int []a2) {
+		boolean flag=false;
 		for(int index=0;index<a1.length;index++) {
-			for(int innerIndex=0;innerIndex<a2.length;innerIndex++) {
-				if(a1[index]==a2[innerIndex]) {
+				if(a1[index]==a2[index]) {
 					System.out.print(a1[index]+" ");
 				}
 			}
-		}
 		}
 	
 	public static void main(String[] args) {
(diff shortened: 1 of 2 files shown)


src/tejashree/CompareTwoArrays.java
```diff
@@ -0,0 +1,29 @@
...
+
+output: 1 7
+*/
+
+package tejashree;
+
+public class CompareTwoArrays {
+
+	public void compareArrays(int []a1, int []a2) {
+		for(int index=0;index<a1.length;index++) {
```
harshvegada, line 16: Hi Tejashree,
Pls update accordingly.

src/tejashree/StringComparison.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class StringComparison {
+
+	public void compareStrings(String s1,String s2) {
+		boolean flag=true;
+		if(s1.length()==s1.length()) {
+		for(int index=0;index<s1.length();index++) {
+			if(!(s1.charAt(index)==s2.charAt(index))) {
+				flag=false;
+			}
```
harshvegada, line 24: put break, to improve code time complexity.

harshvegada (changes requested):
Please update code, based on given comments

src/tejashree/CompareTwoArrays.java
harshvegada, line 16: Earlier code was correct


static void printCommonElement(int a[], int[] b) {
		for (int i = 0; i < a.length; i++) {
			for (int j = 0; j < b.length; j++) {
				if (a[i] == b[j]) {
					System.out.println(a[i]);
					break;
				}
			}
		}
	}

harshvegada (approved):
Code not done yet

PR #951, opened 2020-11-27
Prasad_C_Coding_exam_4_29mins

PrasadCOct2020:
Hi Harsh,

Please review my PR 
image

image

Regards 
Prasad Chitale


src/prasad/Findcommonelefrarray.java
```diff
@@ -0,0 +1,25 @@
+package prasad;
+
+public class Findcommonelefrarray {
+	void scanArray(int[]array1 , int[]array2) {
+		if(array1.length!=array2.length)
+			System.out.println("Given two arrays are not of same length, please retry");
+		else
+			System.out.println("Common Elements:- ");
+		for(int index=0;index<array1.length;index++) {
+			if(array1.length != array2.length)
```
harshvegada, line 10: this condition you already given, then why again here.

harshvegada: Hi Prasad,

src/prasad/Findiftwostrareindentical.java
```diff
@@ -0,0 +1,34 @@
...
+public class Findiftwostrareindentical {
+	
+	boolean scanStr(String input1 , String input2) {
+		boolean flag = false;
+		int count=0;
+		for(int index=0 ; index<input1.length(); index++) {
+			if(input1.length()!=input2.length() || input1.charAt(index)!=input2.charAt(index))
+				break;
+			
+			else if(input1.charAt(index)== input2.charAt(index)) {
```
harshvegada, line 12: you can put condition  != condition and flag = false/true and then check flag value. Accordingly flag value print message,
Enhancement

PrasadCOct2020: Corrected the class please check

src/prasad/Stringandarray/Findcommonelefrarray.java
```diff
@@ -0,0 +1,49 @@
...
+
+input:
+
+int array1 [] = [1, 2, 5, 5, 8, 9, 7, 10]
+int array2 [] = [1, 0, 6, 15, 6, 4, 7, 0]
+
+output: 1 7*/
+
+public class Findcommonelefrarray {
+	void scanArray(int[]array1 , int[]array2) {
```
harshvegada, line 12: Code works fine but time complexity very high
for(){
  while(){
  }
  for(){
   }
}

you can check this code

static void printCommonElement(int a[], int[] b) {
		for (int i = 0; i < a.length; i++) {
			for (int j = 0; j < b.length; j++) {
				if (a[i] == b[j]) {
					System.out.println(a[i]);
					break;
				}
			}
		}
	}

harshvegada (approved):
Update code & acknowledge me!

PR #953, opened 2020-11-27
RohanM_Coding_Exam_4_30mins

RohanMOct2020:
Please review my PR
image
image
image

### DIFF
--- a/src/rohan/IdenticalArrays.java
+++ b/src/rohan/IdenticalArrays.java
@@ -1,0 +1,35 @@
+package rohan;
+
+/*
+ * Program 1: WAP to Find Common Element Between Two Arrays
+
+input: 
+
+int array1 [] = [1, 2, 5, 5, 8, 9, 7, 10]
+int array2 [] = [1, 0, 6, 15, 6, 4, 7, 0]
+
+output: 1 7
+ */
+
+public class IdenticalArrays {
+	
+	public static void main(String[] args) {
+		IdenticalArrays identicalArrays = new IdenticalArrays();
+		int[] numArr1 = {1, 2, 5, 5, 8, 9, 7, 10,21,13};
+		int[] numArr2 = {1, 0, 6, 15, 6, 4, 7, 0,21,3};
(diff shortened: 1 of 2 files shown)


src/rohan/IdenticalArrays.java
```diff
@@ -0,0 +1,35 @@
...
+		IdenticalArrays identicalArrays = new IdenticalArrays();
+		int[] numArr1 = {1, 2, 5, 5, 8, 9, 7, 10,21,13};
+		int[] numArr2 = {1, 0, 6, 15, 6, 4, 7, 0,21,3};
+		identicalArrays.findIdenticalElements(numArr1,numArr2);
+	}
+	
+	void findIdenticalElements(int[] numArr1,int[] numArr2) {
+		String str="";
+		if(numArr1.length == numArr2.length) {
+			for(int index=0;index<numArr1.length;index++) {
```
harshvegada, line 26: Hi Rohan,
you need to compare element in two loop, fortunately both array have same common value on same index, but think about diff index, so that this will not work & also put break once you match the index value.

src/rohan/IdenticalStrings.java
```diff
@@ -0,0 +1,54 @@
...
+	void findIdenticalStrings(String str1,String str2) {
+		boolean flag = true;
+		if(str1.length() == str2.length()) {
+			for(int index=0;index<str1.length();index++) {
+				if(str1.charAt(index) != str2.charAt(index)) {
+						flag = false;
+						break;
+				}
+				else {
+					if(str1.charAt(index) == str2.charAt(index) && Character.isLowerCase(str1.charAt(index))==Character.isLowerCase(str2.charAt(index)) || Character.isUpperCase(str1.charAt(index))==Character.isUpperCase(str2.charAt(index)))
```
harshvegada, line 42: both condition are same if you observe if condition and ese me if condition.
you require only first if condition, rest code not require

harshvegada (changes requested):
work on given comments
Acknowledge me once you update the code!

harshvegada:
Code not updated, merging

PR #956, opened 2020-11-27
RitikaG Coding_Exam_4_35mins

RitikaGOct2020:
image

image


src/ritika/exams/Coding_Exam_4_Pg1.java
```diff
@@ -0,0 +1,34 @@
...
+
+output: 1 7
+*/
+public class Coding_Exam_4_Pg1 {
+
+	static void findCommonElementsBetweenArrays(int[] array1, int[] array2) {
+		if (array1.length == array2.length || array1.length < array2.length) {
+			System.out.println("Common elements:");
+			for (int index = 0; index < array1.length; index++) {
+				if (array1[index] == array2[index])
```
harshvegada, line 17: Hi Ritika,

harshvegada (changes requested):
One change require in array's program, second program code looks good to me!

src/ritika/exams/Coding_Exam_4_Pg1.java
RitikaGOct2020: @harshvegada 
Required changes have been done and code is updated. Time taken to resolve issue is approx 15mins.
Thanks for reviewing!

harshvegada (approved):

harshvegada:
Good to merge!

PR #958, opened 2020-11-27
ShrutiC_Coding_Exam4_25Mins

ShrutiCOct2020:
image

### DIFF
--- a/src/shrutiC/CodingExam4.java
+++ b/src/shrutiC/CodingExam4.java
@@ -1,0 +1,44 @@
+package shrutiC;
+
+public class CodingExam4 {
+
+	void findSimilarNumInArray(int[] array1, int[] array2) {
+
+		String similarNum = "";
+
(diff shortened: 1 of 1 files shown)


src/shrutiC/CodingExam4.java
```diff
@@ -0,0 +1,44 @@
...
+public class CodingExam4 {
+
+	void findSimilarNumInArray(int[] array1, int[] array2) {
+
+		String similarNum = "";
+
+		for (int i = 0; i < array1.length; i++) {
+
+			if (array1[i] == array2[i])
+				similarNum = similarNum + " " + Integer.toString(array1[i]);
```
harshvegada, line 12: Hi Shruti,

harshvegada (changes requested):
Where is 2nd program
Please add missing program

harshvegada:
code not updated

PR #961, opened 2020-11-27
Suresh exam 4_20 mins

sureshchukka:
Hi Harsh,

Please Review my Assignment - Coding Exam4.

Thanks & Regards,
Suresh Chukka.

### DIFF
--- a/src/suresh/StringsIdenticalwithoutArray.java
+++ b/src/suresh/StringsIdenticalwithoutArray.java
@@ -1,6 +1,6 @@
 package suresh;
 
-public class StringsIdentical {
+public class StringsIdenticalwithoutArray {
 
 		void identicalStringFinder() {
 			/*String s1 = "technoCredits";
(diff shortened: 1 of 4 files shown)


src/suresh/StringsIdenticalwithoutArray.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class StringsIdenticalwithoutArray {
+
+		void identicalStringFinder() {
+			/*String s1 = "technoCredits";
+			String s2 = "teChnoCreditS";*/
+			
+			String s1 = "pune";
+			String s2 = "pune";
+			if(s1==s2) {
```
harshvegada, line 11: make parametrized method
you need to use for loop and compare single char with another string char

harshvegada (changes requested):
work on comments
update code and let me know

harshvegada (approved):
code not updated as per the comments

PR #964, opened 2020-11-27
Coding_Exam4_MrunalY

MrunalYOct2020:
image
image


src/mrunal/program1.java
```diff
@@ -0,0 +1,42 @@
...
+	
+	void getIdenticalNumbers(int array1 []  , int array2 [] ) {
+		for(int index1=0; index1<array1.length; index1++) {
+			int num=0 ; 
+			int num1=array1[index1];
+			for(int index2=0 ; index2<array2.length ; index2++) {
+				int num2=array2[index2];
+				if(num1==num2) {
+					num=num1;
+					System.out.println("Identical elements are :" + num);
```
harshvegada, line 23: you need to put break here,

harshvegada (approved):
Code looks Good to me!
Match require output!
you not mentioned time

PR #965, opened 2020-11-27
YogitaP_Coding_Exam_4_45min

YogitaPOct2020:
image

image

### DIFF
--- a/src/yogita/FindIdenticalElement.java
+++ b/src/yogita/FindIdenticalElement.java
@@ -1,0 +1,40 @@
+package yogita;
+
+import java.util.Arrays;
+
+/*
+ * Program 1: WAP to Find Common Element Between Two Arrays
+
+
+input: 
+
+int array1 [] = [1, 2, 5, 5, 8, 9, 7, 10]
+int array2 [] = [1, 0, 6, 15, 6, 4, 7, 0]
+
+output: 1 7
+
+*/
+public class FindIdenticalElement {
+	void identicalElementFRomArray(int[] array1, int[] array2) {
+		System.out.println("Array1 is:" + Arrays.toString(array1));
+		System.out.println("Array2 is:" + Arrays.toString(array2));
+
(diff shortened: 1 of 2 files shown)


src/yogita/FindIdenticalElement.java
```diff
@@ -0,0 +1,40 @@
...
+	void identicalElementFRomArray(int[] array1, int[] array2) {
+		System.out.println("Array1 is:" + Arrays.toString(array1));
+		System.out.println("Array2 is:" + Arrays.toString(array2));
+
+		System.out.println("Identical Elemets From Array Are:");
+		for (int i = 0; i < array1.length; i++) {
+			boolean flag = false;
+			for (int j = 0; j < array2.length; j++) {
+				if (array1[i] == array2[j])
+					flag = true;
```
harshvegada, line 27: just need to put break

src/yogita/FindIdenticalString.java
```diff
@@ -0,0 +1,47 @@
...
+public class FindIdenticalString {
+	void findIdenticalString(String str1, String str2) {
+
+		boolean flag = false;
+		if (str1.length() == str2.length()) {
+			for (int i = 0; i < str1.length(); i++) {
+				char ch1 = str1.charAt(i);
+				char ch2 = str2.charAt(i);
+				if (ch1 != ch2)
+					flag = true;
```
harshvegada, line 24: Hi ruby, 
here you should put break; once you know that char is matching then no more iterations are require

harshvegada (approved):

src/yogita/FindIdenticalElement.java
YogitaPOct2020: Is it necessary to use break with flag everytime  ?

src/yogita/FindIdenticalString.java
YogitaPOct2020: Ok.. got it..

PR #966, opened 2020-11-27
RubyK_CodingExam_4

rubykoct2020:
image

image

### DIFF
--- a/src/ruby/Exam4Program1.java
+++ b/src/ruby/Exam4Program1.java
@@ -1,0 +1,33 @@
+package ruby;
+/*Program 1: WAP to Find Common Element Between Two Arrays
+
+input: 
+
+int array1 [] = [1, 2, 5, 5, 8, 9, 7, 10]
+int array2 [] = [1, 0, 6, 15, 6, 4, 7, 0]
+
+output: 1 7*/
+
+public class Exam4Program1 {
+	void findCommonElements(int[] array1, int[] array2) {
+		boolean flag = false;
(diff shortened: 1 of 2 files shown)


src/ruby/Exam4Program1.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class Exam4Program1 {
+	void findCommonElements(int[] array1, int[] array2) {
+		boolean flag = false;
+		for (int i = 0; i < array1.length; i++) {
+			if (array1.length == array2.length) {
+				flag = true;
+			if (array1[i] == array2[i]) {
+					System.out.print(array1[i]+" ");
+				}
```
harshvegada, line 19: just need to put break, as you get matching value in second array

src/ruby/Exam4Program2.java
```diff
@@ -0,0 +1,47 @@
...
+public class Exam4Program2 {
+	void CompareString(String s1, String s2) {
+		boolean flag = false;
+		int count = 0;
+		for (int i = 0; i < s1.length(); i++) {
+			if (s1.length() == s2.length()) {
+				flag = true;
+				if (s1.charAt(i) != s2.charAt(i)) {
+					count++;
+				}
```
harshvegada, line 25: you should  put break, as you know now char is not matching then you can break the loop

harshvegada (approved):
Code looks Good to me!
Match require output!
you not mentioned time, hoping you done in time limit

PR #967, opened 2020-11-27
Nitin_Coding_Exam_4_#40min

NitinJOct2020:
Program 1 -
image

Program 2 -
image

### DIFF
--- a/src/nitin/Exam/Exam4_1.java
+++ b/src/nitin/Exam/Exam4_1.java
@@ -1,0 +1,30 @@
+package nitin.Exam;
+import java.util.Arrays;
+/* Exam 4 - Program 1: WAP to Find Common Element Between Two Arrays
+input: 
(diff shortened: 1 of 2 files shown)


src/nitin/Exam/Exam4_1.java
```diff
@@ -0,0 +1,30 @@
...
+		System.out.print("Common Element Between Two Arrays --> ");
+		for(int index=0;index<arr1.length;index++) {
+			int num=0;
+			int temp=arr1[index];
+			for(int inner=0;inner<arr2.length;inner++) {
+				int temp1=arr2[inner];
+				if(temp==temp1) {
+					num=temp;
+					System.out.print(" "+num);
+				}
```
harshvegada, line 21: Hi Nitin,
You need to put break, because once you get your match value in second array then you won't iterate loop

harshvegada (approved):
Very Good catch, nitin
probably you cover almost possible scenario.

PR #968, opened 2020-11-27
RohitR_Exam_4

RohitROct2020:


src/rohit/Commonelement.java
```diff
@@ -0,0 +1,32 @@
...
+public class Commonelement {
+
+	void arraycommonelement(int[] array1, int[] array2) {
+		System.out.println("Common elements are: ");
+		for (int i = 0; i < array1.length; i++) {
+			for (int j = 0; j < array2.length; j++) {
+				if (array1[i] == (array2[j])) {
+
+					System.out.println((array1[i]));
+				}
```
harshvegada, line 20: You just need to put break, once you match the element value in second array

harshvegada (approved):
Code looks Good to me!
Match require output!
you not mentioned time, hope you completed with in time limit

src/rohit/Commonelement.java
RohitROct2020: ok will take care of that from next time took 40 mins time to complete it, thanks for your comments.

PR #969, opened 2020-11-27
Java coding exam 4

NagendraTOct2020:


src/nagendra/CodingExam4.java
```diff
@@ -0,0 +1,30 @@
...
+
+
+public class CodingExam4 {
+
+	
+	void getCommonElement(int[] input,int[] input2) {
+		for(int i=0;i<input.length;i++){
+			for(int j=0;j<input2.length;j++){
+				if(input[i]==input2[j]){
+					System.out.println("Common :"+input[i]);
```
harshvegada, line 18: Just need to put break, once you get element value then you can stop iteration on second array,

src/nagendra/CodingExam4A.java
```diff
@@ -0,0 +1,42 @@
...
+*/
+
+public class CodingExam4A {
+	
+	void getIdenticalString(String userInput, String input) {
+		boolean flag = true;
+		char[] input1=userInput.toLowerCase().toCharArray();
+		char[] input2=input.toLowerCase().toCharArray();
+       for(int i=0;i<input1.length;i++){
+    	   for(int j=i;j<=i;j++){
```
harshvegada, line 21: you can handle in single loop,
and need to use charAt() method to compare char value, if not match then update flag value and break the loop,

harshvegada (changes requested):
Work on comments for 2nd String program

src/nagendra/CodingExam4.java
NagendraTOct2020: Updated as per comment

harshvegada: Thanks Nagendra

src/nagendra/CodingExam4A.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class CodingExam4A {
+
+	void getIdenticalString(String userInput, String input) {
+		boolean flag = true;
+		String input1 = userInput.toLowerCase();
+		String input2 = input.toLowerCase();
+		for (int i = 0; i < input1.length(); i++) {
+			if (input1.charAt(i) != input2.charAt(i))
+				flag = false;
```
harshvegada, line 22: here also need to put break;

harshvegada (approved):
Good to merge!

PR #970, opened 2020-11-27
KajolC_Coding_Exam_4-40mins

KajolCOct2020:
CodingExam-4_Program1-FindCommonElementsBetweenTwoArrays

CodingExam-4_Program2-FindIdenticalStrings

### DIFF
--- a/src/kajol/FindCommonElementsBetweenTwoArrays.java
+++ b/src/kajol/FindCommonElementsBetweenTwoArrays.java
@@ -1,0 +1,37 @@
+/*Program 1: WAP to Find Common Element Between Two Arrays
+
+input: 
+
+int array1 [] = [1, 2, 5, 5, 8, 9, 7, 10]
+int array2 [] = [1, 0, 6, 15, 6, 4, 7, 0]
+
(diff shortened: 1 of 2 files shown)


src/kajol/FindCommonElementsBetweenTwoArrays.java
```diff
@@ -0,0 +1,37 @@
...
+public class FindCommonElementsBetweenTwoArrays {
+	 int cmnelements=0;
+	void findCommonElements(int array1[],int array2[]) {
+		for(int index1=0;index1<array1.length;index1++) {
+			for(int index2=0;index2<array2.length;index2++) {
+				if(array1[index1]==array2[index2]) {
+					cmnelements++;
+					if(cmnelements==1)
+						System.out.print("Common Elements are : ");
+					System.out.print(array1[index1]+" ");
```
harshvegada, line 22: Just one small change, you should put break, as you get your match then why we need to iterate on second array right

harshvegada (approved):
Code looks Good to me!
Match require output!
Need to improve time ! Keep practicing

PR #971, opened 2020-11-27
AnjaliS_Coding_Exam4

AnjaliSoct2020:
image

image

### DIFF
--- a/src/anjaliS/PrintSameNumberFromArray.java
+++ b/src/anjaliS/PrintSameNumberFromArray.java
@@ -1,0 +1,21 @@
+package anjaliS;
+
+public class PrintSameNumberFromArray {
+	void displaySameNum(int[] arr1, int[] arr2) {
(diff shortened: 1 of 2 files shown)


src/anjaliS/PrintSameNumberFromArray.java
```diff
@@ -0,0 +1,21 @@
+package anjaliS;
+
+public class PrintSameNumberFromArray {
+	void displaySameNum(int[] arr1, int[] arr2) {
+		for (int i = 0; i < arr1.length; i++) {
+			for (int j = 0; j < arr2.length; j++) {
+				if (arr2[j] == arr1[i])
+					System.out.println(+arr1[i]);
```
harshvegada, line 8: Just one small change, you should put break, as you get your match then why we need to iterate on second array right

harshvegada (approved):
Code looks Good to me!
Match require output!
you not mentioned time, hope you completed exam with in time!

PR #982, opened 2020-11-27
Sadhana_Coding_Exam_4

SadhanaPOct2020:
@harshvegada @KrishnaKTechnocredits please review my PR

### DIFF
--- a/src/sadhana/CommonElementsBitweenArray.java
+++ b/src/sadhana/CommonElementsBitweenArray.java
@@ -1,0 +1,25 @@
+package sadhana;
+
+//Coding exam_4
+
+public class CommonElementsBitweenArray {
+
+	void printData(int[] arr1, int[] arr2) {
+		for (int i = 0; i < arr1.length; i++) {
+			for (int j = 0; j < arr2.length; j++) {
+				if (arr2[j] == arr1[i])
+					System.out.println(+arr1[i]);
+			}
+		}
+	}
+
+	public static void main(String[] args) {
+		CommonElementsBitweenArray commonElementsBitweenArray = new CommonElementsBitweenArray();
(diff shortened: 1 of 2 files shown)


src/sadhana/CommonElementsBitweenArray.java
```diff
@@ -0,0 +1,25 @@
...
+//Coding exam_4
+
+public class CommonElementsBitweenArray {
+
+	void printData(int[] arr1, int[] arr2) {
+		for (int i = 0; i < arr1.length; i++) {
+			for (int j = 0; j < arr2.length; j++) {
+				if (arr2[j] == arr1[i])
+					System.out.println(+arr1[i]);
+			}
```
harshvegada, line 12: you need to break, as once we get match element value then we don't want to check for other element in second array

src/sadhana/IdenticalNotIdenticalArray.java
```diff
@@ -0,0 +1,33 @@
...
+//Coding exam_4
+
+public class IdenticalNotIdenticalArray {
+
+	void printData(String msg1, String msg2) {
+		boolean flag = false;
+		if (msg1.length() == msg2.length()) {
+			for (int i = 0; i < msg1.length(); i++) {
+				if (msg1.charAt(i) != msg2.charAt(i))
+					flag = true;
```
harshvegada, line 12: same as above comment, need to put break, as you know for this char value's are not matching mean string are not identical, so we can update flag value and break the loop

harshvegada (approved):
Code looks good to me @SadhanaPOct2020 
Good to merge!

harshvegada (approved):
Code looks good to me!
Good to merge

PR #985, opened 2020-11-27
BrijeshG_Assignment_29

BrijeshGOct2020:
image

### DIFF
--- a/src/brijesh/assignment29/ArraySecondMax.java
+++ b/src/brijesh/assignment29/ArraySecondMax.java
@@ -7,29 +7,36 @@ Find second maximum from array without sorting the array.
 */
 
 public class ArraySecondMax {
-	
-	public int findMax(int[] input) {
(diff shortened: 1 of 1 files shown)


harshvegada (changes requested):
we can handle in single loop only right. in your case time complexity is 2n

check this code, I m sure you get my idea, please update code n acknowledge me.
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/nandini/SecondMaxNumber.java

BrijeshGOct2020:
Hi @harshvegada 

Updated code. Handling negative and all corner cases as well. Thanks
image

BrijeshGOct2020:
Hi @harshvegada,

Thanks
Updated code with all corner cases and negative number
image

harshvegada (approved):
Code looks good to me! 
very optimized code!
Good to merge

PR #989, opened 2020-11-27
VaibhavJ_Assignment_29

VaibhavJOct2020:
image

### DIFF
--- a/src/vaibhav/SecondMax.java
+++ b/src/vaibhav/SecondMax.java
@@ -1,0 +1,28 @@
+package vaibhav;
+
+/*Assignment - 29 : 26th Nov'2020 
+Find second maximum from array without sorting the array.*/
+
+public class SecondMax {
+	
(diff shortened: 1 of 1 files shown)


src/vaibhav/SecondMax.java
```diff
@@ -0,0 +1,28 @@
+package vaibhav;
+
+/*Assignment - 29 : 26th Nov'2020 
+Find second maximum from array without sorting the array.*/
+
+public class SecondMax {
+	
+	public static void getSecondElement(int[] array1) {
+		int len = array1.length;
+		for (int index = 0 ; index < len;index++) {
```
harshvegada, line 10: Sorting logic is correct but in sorting you are using two loops here,
you can hanld in single loop
Check this code out
https://github.com/KrishnaKTechnocredits/JAVATechnoOct2020/blob/main/src/nandini/SecondMaxNumber.java

harshvegada (approved):

PR #992, opened 2020-11-28
Tejashree t assignment29

TejashreeTOct2020:
Please review 
image

### DIFF
--- a/src/tejashree/Assignment29.java
+++ b/src/tejashree/Assignment29.java
@@ -1,0 +1,24 @@
+package tejashree;
+
+public class Assignment29 {
+	void findSecondMaxInArray(int[] arr) {
+		int max=arr[0];
+		int secondMax=0;		
+		for(int index=2;index<arr.length;index++) {
+			if(arr[index]>max) {
+				secondMax=max;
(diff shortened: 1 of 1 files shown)


src/tejashree/Assignment29.java
```diff
@@ -0,0 +1,24 @@
+package tejashree;
+
+public class Assignment29 {
+	void findSecondMaxInArray(int[] arr) {
+		int max=arr[0];
+		int secondMax=0;		
+		for(int index=0;index<arr.length;index++) {
```
harshvegada, line 7: you can start index from 2 as 
int max=arr[0];
int secondMax=arr[1];

src/tejashree/Assignment29.java
```diff
@@ -0,0 +1,24 @@
...
+			if(arr[index]>max) {
+				secondMax=max;
+				max=arr[index];
+			}
+			else if(arr[index]>secondMax) {
```
harshvegada, line 12: you should also check

harshvegada (approved):
Code looks good to me!
Good to merge

PR #993, opened 2020-11-28
MrunalY_Assignment23

MrunalYOct2020:
image

### DIFF
--- a/src/mrunal/PrimeNumber_Addition_Average.java
+++ b/src/mrunal/PrimeNumber_Addition_Average.java
@@ -1,0 +1,36 @@
+package mrunal;
+import java.util.Arrays;
+
+public class PrimeNumber_Addition_Average {
+
+	void primeNumber(int startRange, int endRange) {
+		int sum = 0;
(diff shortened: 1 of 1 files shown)


src/mrunal/PrimeNumber_Addition_Average.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Arrays;
+
+public class PrimeNumber_Addition_Average {
+
+	void primeNumber(int startRange, int endRange) {
+		int sum = 0;
+		int countPrimeNumber = 0;
+		for(int index = startRange; index <= endRange; index++) {
+			int count = 0;
+			for(int innerIndex = startRange; innerIndex <= index; innerIndex++) {
```
harshvegada, line 11: this startRange always starts with 2 & range should be Math.sqrt(index);

harshvegada (approved):
Please keep in mind logically why from 2 for prime number, this is most fav question in interview,
take of this next time

PR #1005, opened 2020-11-28
CodingExam_4

NandiniSOct2020:
Added 2 java files and it took around total 15 mins to complete it. Please review.

### DIFF
--- a/src/nandini/CodingExams/CodingExam4.java
+++ b/src/nandini/CodingExams/CodingExam4.java
@@ -1,0 +1,25 @@
+package nandini.CodingExams;
+//Program 1: WAP to Find Common Element Between Two Arrays
+
+public class CodingExam4 {
+
+	void findDuplicateNumber(int[] input1,int[] input2) {
+		int count =0;
+		for(int i=0; i<input1.length; i++) {
+			for(int j=0; j<input2.length; j++) {
+				if(input1[i] == input2[j]) 
+					count++;
+			}
+			if (count > 1)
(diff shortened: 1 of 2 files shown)


src/nandini/CodingExams/CodingExam4.java
```diff
@@ -0,0 +1,25 @@
...
+//Program 1: WAP to Find Common Element Between Two Arrays
+
+public class CodingExam4 {
+
+	void findDuplicateNumber(int[] input1,int[] input2) {
+		int count =0;
+		for(int i=0; i<input1.length; i++) {
+			for(int j=0; j<input2.length; j++) {
+				if(input1[i] == input2[j]) 
+					count++;
```
harshvegada, line 11: once we found duplicate number then you should put break, you don't need count variable as well, you directly print the value in if condition

src/nandini/CodingExams/CodingExam4_Prog2.java
```diff
@@ -0,0 +1,31 @@
+package nandini.CodingExams;
+//WAP to find two String are identical or not
+public class CodingExam4_Prog2 {
+	
+	void notIdenticalStrings(String str1, String str2) {
+				if((str1.equals(str2)))
```
harshvegada, line 6: As i mention you will not use .equlas method

harshvegada (approved):
Code looks good but in String program this is not expected,
you need to use charAt(i) method

PR #1023, opened 2020-11-29
ShrutiS_Assignment_29

ShrutiSOct2020:
image


src/shrutiS/assignment29/SecondMaxNumber.java
```diff
@@ -0,0 +1,33 @@
...
+	void secondMaximumNumber(int arr[]) {
+
+		int max = arr[0];
+		int secondMax = arr[1];
+		for (int index = 0; index < arr.length; index++) {
+			for (int innerIndex = 2; innerIndex < arr.length; innerIndex++) {
+				if (arr[innerIndex] > max) {
+					secondMax = max;
+					max = arr[innerIndex];
+				} else if (arr[innerIndex] > secondMax)
```
harshvegada, line 20: you should one extra condition for max number is like

harshvegada (approved):
rest code looks good to me!
Logic looks correct!

PR #1024, opened 2020-11-29
ShrutiC_Assignment_23

ShrutiCOct2020:
image

### DIFF
--- a/src/shrutiC/OperationOnPrimeNumbers.java
+++ b/src/shrutiC/OperationOnPrimeNumbers.java
@@ -1,0 +1,79 @@
+package shrutiC;
+
+public class OperationOnPrimeNumbers {
+
+	int primeNumSum;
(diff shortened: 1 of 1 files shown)


src/shrutiC/OperationOnPrimeNumbers.java
```diff
@@ -0,0 +1,79 @@
+package shrutiC;
+
+public class OperationOnPrimeNumbers {
+
+	int primeNumSum;
+	int totalNumOfPrimeNos;
+
+	void finSumOfPrimeNumbers(int startNum, int endNum) {
+		int sum = 0;
+		for (int number = startNum; number <= endNum; number++) {
```
harshvegada, line 10: when ever you write prime number code, make sure you are define range 2, reason we already seen in class

harshvegada (approved):
Code looks good to me!
Good to merge

PR #1028, opened 2020-11-29
AnjaliS_Assignment30

AnjaliSoct2020:
image

image

### DIFF
--- a/src/anjaliS/DuplicateNumberFromArray.java
+++ b/src/anjaliS/DuplicateNumberFromArray.java
@@ -1,0 +1,21 @@
+package anjaliS;
+
+public class DuplicateNumberFromArray {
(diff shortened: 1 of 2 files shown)


harshvegada (approved):
Code looks good to me!
Good to merge

assume there are 3 same number in array, so in that case this code will print twice. so to overcome this you should use break & need to update array's value.

AnjaliSoct2020:
> Code looks good to me!
> Good to merge
> 
> assume there are 3 same number in array, so in that case this code will print twice. so to overcome this you should use break & need to update array's value.

Tried with 3same number in array, & in that case also code is printed value once only.. with using break.. sending you screenshot on skype & please check and let me knw

PR #1048, opened 2020-11-30
NandiniS_Assignment_30

NandiniSOct2020:
Added two java files.Please review.

### DIFF
--- a/src/nandini/MissingNumber.java
+++ b/src/nandini/MissingNumber.java
@@ -1,0 +1,22 @@
+package nandini;
+
+public class MissingNumber {
+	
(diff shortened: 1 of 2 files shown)


src/nandini/MissingNumber.java
```diff
@@ -0,0 +1,22 @@
+package nandini;
+
+public class MissingNumber {
+	
+	void findMissingNumber(int[] arr) {
+		int expNoOfElements = arr.length + 1;
+		int totalsum = expNoOfElements * (expNoOfElements + 1) / 2;
```
harshvegada, line 7: this is what m expecting

PR #1055, opened 2020-11-30
RitikaG_Assignment_30_All Programs

RitikaGOct2020:
image

image

### DIFF
--- a/src/ritika/Assignments/Asgmt_30_Pg1.java
+++ b/src/ritika/Assignments/Asgmt_30_Pg1.java
@@ -1,0 +1,33 @@
+package ritika.Assignments;
+
+/*Assignment 30 : 29th Nov 2020
+
+Program 1: Find the missing number in integer array of 1 to 10?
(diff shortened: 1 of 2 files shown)


src/ritika/Assignments/Asgmt_30_Pg1.java
```diff
@@ -0,0 +1,33 @@
...
+
+Program 1: Find the missing number in integer array of 1 to 10?
+Hint: You have given an integer array which contains numbers from 1 to 10 but exactly one number is missing, you need to write a Java program to find that missing number in an array.
+
+*/
+public class Asgmt_30_Pg1 {
+
+	static void findMissingNumber(int[] input) {
+
+		int[] matchArray = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
```
harshvegada, line 13: this logic is also correct, with two loops, you can manage in single loop as well,
you can take on sum variable named expected Sum assume in this case it's 55 & you need to do sum of array element & names array sum as actual Sum, then perform minus operation, you will get missing number.

harshvegada (approved):
rest logic looks good to me!
Good to merge!

PR #1070, opened 2020-12-01
PriyaM_Assignment_30

PriyaMOct2020:
image

image

Please review my assignment 30

### DIFF
--- a/src/priya/DuplicateNumberInArray.java
+++ b/src/priya/DuplicateNumberInArray.java
@@ -1,0 +1,24 @@
+//Program 2:  Find duplicate number on Integer array in Java? [I know you have done this before, practice it one more time please]
+//Hint: There is exactly one number is repeated in the array. You need to write a program to
+//find that duplicate number.
+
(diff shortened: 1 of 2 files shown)


src/priya/DuplicateNumberInArray.java
```diff
@@ -0,0 +1,24 @@
...
+
+package priya;
+
+public class DuplicateNumberInArray {
+
+	void displayDuplicateNum(int arr[]) {
+		for (int i = 0; i < arr.length; i++) {
+			for (int j = i + 1; j < arr.length; j++) {
+				if (arr[j] == arr[i]) {
+					System.out.println("Duplicate number in array is:" + arr[j]);
```
harshvegada, line 13: you should break here

harshvegada: because once match the element value then why go for further iterations.

harshvegada (approved):
Code looks good to me!
Good to merge

PR #1073, opened 2020-12-01
Suresh asgnmt 26

sureshchukka:
Hi Krishna/harsh,

Could you please Review my Assignment-26.
Sorry for the delay... 

Thanks & Regards,
Suresh Chukka.
image

### DIFF
--- a/src/suresh/DuplicateCheck.java
+++ b/src/suresh/DuplicateCheck.java
@@ -1,0 +1,24 @@
+package suresh;
+
+public class DuplicateCheck {
+	
+	void duplicates(int[] names) {
+		 for (int i = 0; i < names.length; i++) {
+		     for (int j = i + 1 ; j < names.length; j++) {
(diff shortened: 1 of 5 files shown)


src/suresh/DuplicateCheck.java
```diff
@@ -0,0 +1,24 @@
+package suresh;
+
+public class DuplicateCheck {
+	
+	void duplicates(int[] names) {
+		 for (int i = 0; i < names.length; i++) {
+		     for (int j = i + 1 ; j < names.length; j++) {
+		          if (names[i]==(names[j])) {
```
harshvegada, line 8: this code will not work for below input
1,2,3,1,2,1,4,6,5,5,2

harshvegada (changes requested):
rest programs looks good, update code based on comments and try to debug where you missed to put condition, once you push code again then let me know

PR #1074, opened 2020-12-01
ShrutiS_Assignment_30

ShrutiSOct2020:
Program 1

image

Program 2

image


src/shrutiS/assignment30/DuplicateNumber.java
```diff
@@ -0,0 +1,36 @@
...
+		for (int index = 0; index < input.length; index++) {
+			System.out.print(input[index] + " ");
+		}
+	}
+	
+	void findDuplicateNumber(int[] input) {
+		for(int index =0; index<input.length; index++) {
+			for(int innerIndex = index+1; innerIndex<input.length; innerIndex++) {
+				if(input[index] == input[innerIndex])
+					System.out.println("\nDuplicate number is "+input[index]);
```
harshvegada, line 21: you should put break, as once you get match element then you don't require to iterate

harshvegada (approved):
Code Matches expected output!
Logic looks good to me!

PR #1084, opened 2020-12-02
ShrutiC_Assignment_29

ShrutiCOct2020:
image

### DIFF
--- a/src/shrutiC/FindSecondMaxNum.java
+++ b/src/shrutiC/FindSecondMaxNum.java
@@ -1,0 +1,29 @@
+package shrutiC;
+
+public class FindSecondMaxNum {
+
+	void getSecoundMaxNumInArray(int arrayNum[]) {
+		int max = arrayNum[0];
+		int secondMaxNum = 0;
+		System.out.print("Input array : {" + max);
+		for (int i = 1; i < arrayNum.length; i++) {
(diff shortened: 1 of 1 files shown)


src/shrutiC/FindSecondMaxNum.java
```diff
@@ -0,0 +1,29 @@
+package shrutiC;
+
+public class FindSecondMaxNum {
+
+	void getSecoundMaxNumInArray(int arrayNum[]) {
+		int max = arrayNum[0];
+		int secondMaxNum = 0;
```
harshvegada, line 7: you can start this from a[1] & then for loop from 2

src/shrutiC/FindSecondMaxNum.java
```diff
@@ -0,0 +1,29 @@
...
+		System.out.print("Input array : {" + max);
+		for (int i = 1; i < arrayNum.length; i++) {
+			System.out.print(" " + arrayNum[i]);
+			if (arrayNum[i] > max) {
+				secondMaxNum = max;
+				max = arrayNum[i];
+			}
```
harshvegada, line 14: you need to put else if() condition as well

harshvegada (approved):
One condition needs to be updated,
Keep it up!
take of this next time

PR #1135, opened 2020-12-04
PrasadC_Assignment_34

PrasadCOct2020:
Hi Both,
Please review my PR.

Program 1: O/P
image

Program 2: O/P
image

### DIFF
--- a/src/prasad/Stringandarray/IsStringPalindrome.java
+++ b/src/prasad/Stringandarray/IsStringPalindrome.java
@@ -1,0 +1,45 @@
+package prasad.Stringandarray;
+import java.util.Scanner;
+/*Program 1:  verify string is palindrom or not.
+input : naman
+output : naman is a palindrom string.
+input : techno
+output : techno is not a palindrom string.*/
+
+public class IsStringPalindrome {
+	String output;
+	boolean isPalindrome(String input) {
+		boolean flag = false;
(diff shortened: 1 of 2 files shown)


src/prasad/Stringandarray/IsStringPalindrome.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class IsStringPalindrome {
+	String output;
+	boolean isPalindrome(String input) {
+		boolean flag = false;
+		for(int index=input.length()-1;index>=0;index--) {
+			char ch = input.charAt(index);
+			output+=ch;
+		}
+		output=output.substring(4);
```
harshvegada, line 17: what is use of 4 here?

harshvegada (commented):
Hi @PrasadCOct2020 
Could please tell me use of 4 I didn't get this

src/prasad/Stringandarray/IsStringPalindrome.java
PrasadCOct2020: Hi Harsh,
Since I am concatenating ch with instance String variable which contains default value NULL. So after I have added all the chars from input word to instance String variable. I am using substring to ignore NULL so that in output I only get chars from I/P.
output=output.substring(4) will always give us actual string output minus NULL.

output = NULLnaman
output=output.substring(4) ==> naman

Thanks!

PR #1170, opened 2020-12-05
Suresh_Assignment_34

sureshchukka:
Hi Harsh/Krishna,

Could you please review the Assignment -34.

Thanks & Regards,
Suresh Chukka.
image

image

### DIFF
--- a/src/suresh/StringPalindromeCheck.java
+++ b/src/suresh/StringPalindromeCheck.java
@@ -1,0 +1,25 @@
+package suresh;
+// verify string is palindrom or not.
+public class StringPalindromeCheck {
+	static boolean palindromestring(String str) {
+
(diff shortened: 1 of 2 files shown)


src/suresh/StringPalindromeCheck.java
```diff
@@ -0,0 +1,25 @@
+package suresh;
+// verify string is palindrom or not.
+public class StringPalindromeCheck {
+	static boolean palindromestring(String str) {
+
+		int i = 0, j = str.length() - 1;
+		while (i < j) {
```
harshvegada, line 7: Good to see while loop here

harshvegada (approved):
code logic looks good to me!
Good to merge
Always think diff-diff possible solution

PR #1177, opened 2020-12-05
VishakhaB_Asgnmt_31

VishakhaBOct2020:
image
image

### DIFF
--- a/src/vishakha/Asgnmt31_SumOfNos_ConvertCase.java
+++ b/src/vishakha/Asgnmt31_SumOfNos_ConvertCase.java
@@ -1,0 +1,37 @@
+/*
+Do Sum of all numbers and Convert Uppercase into lowercase, lowercase into uppercase using ASCII.
+Given String = "Te1ch8n9oC6reDi5TS1"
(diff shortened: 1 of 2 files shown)


src/vishakha/Asgnmt31_SumOfNos_ConvertCase.java
```diff
@@ -0,0 +1,37 @@
...
+		int sum=0;
+		boolean flag=false;
+		for(int index=0; index<string.length(); index++) {
+			char character=string.charAt(index);
+			if(character>=65 && character<=90)
+				newString+=(char)(character+32);
+			else if(character>=97 && character<=122)
+				newString+=(char)(character-32);
+			else if(character>=48 && character <=57) {
+				sum+=Integer.parseInt(String.valueOf(character));
```
harshvegada, line 22: This line looks good to me, but here you can also use Character.getNumericValue(character)
this makes less number of method calling & also shows that we know method of character class

harshvegada (approved):
Code looks good to me!
Good to merge

PR #1184, opened 2020-12-06
VishakhaB_Asgnmt_32

VishakhaBOct2020:
image


harshvegada (commented):
Please add correct third program
Program 3:  write a program having below methods.
1) write a method which take one character and return its ascii value.
2) write a method which take one ascii and return character.
3) write a method to print ascii values for uppercase letters [A-Z]. 
A -> 65, B - 66....
4) write a method to print ascii values for lowercase letter [a-z].
a -> 97, b -> 98

PR #1188, opened 2020-12-07
ShrutiC_Assignment_33

ShrutiCOct2020:
image

image

### DIFF
--- a/src/shrutiC/ArmstrongNum.java
+++ b/src/shrutiC/ArmstrongNum.java
@@ -1,0 +1,28 @@
+package shrutiC;
+
+public class ArmstrongNum {
+
+	void isArmstrongNum(int number) {
+		int num = 0;
+		int sumOfCubes = 0;
+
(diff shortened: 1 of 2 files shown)


src/shrutiC/ArmstrongNum.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class ArmstrongNum {
+
+	void isArmstrongNum(int number) {
+		int num = 0;
+		int sumOfCubes = 0;
+
+		String string = Integer.toString(number);
+		for (int i = 0; i < string.length(); i++) {
+			num = Integer.parseInt(Character.toString(string.charAt(i)));
```
harshvegada, line 11: you can use Character.getNumericValue(ch) as second option

but you here interviewer expect like you code for int number only,

harshvegada (approved):
Code looks good to me shruti!
Keep practicing 
Good to merge!

PR #1190, opened 2020-12-07
RitikaG_Assignment_36_All programs

RitikaGOct2020:
image

### DIFF
--- a/src/ritika/Assignments/Asgmt_36.java
+++ b/src/ritika/Assignments/Asgmt_36.java
@@ -1,0 +1,77 @@
+package ritika.Assignments;
+
+import java.util.Arrays;
+
+/*Assignment : 36 [6th Dec'2020]
+
+Program 1: Find union of two arrays.
+Input: arr1 -> {10.45, 14.0, 18.35, 88.88, 54.10, 18.35}
+arr2 -> {17.20, 13.30, 10.45, 18.35, 84.33, 13.30}
+
+Output: 10.45, 14.0, 18.35, 88.88, 54.10, 17.20, 13.30, 84.33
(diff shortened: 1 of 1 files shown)


src/ritika/Assignments/Asgmt_36.java
```diff
@@ -0,0 +1,77 @@
...
+		return outputIntersection;
+	}
+		
+	public static void main(String[] args) {
+		double arr1[] = { 10.45, 14.0, 18.35, 88.88, 54.10, 18.35 };
+		double arr2[] = { 17.20, 13.30, 10.45, 18.35, 84.33, 13.30 };
+		Asgmt_36 program1 = new Asgmt_36();
+		System.out.println("Union of given 2 arrays is: " + Arrays.toString(program1.getUnionOfArray(arr1, arr2)));
+		System.out.println();
+		Asgmt_36 program2 = new Asgmt_36();
```
harshvegada, line 74: You don't need to create object 2nd time you can call it on same reference number as well which is program1 in your case.

src/ritika/Assignments/Asgmt_36.java
```diff
@@ -0,0 +1,77 @@
...
+
+Output: 10.45, 18.35
+*/
+public class Asgmt_36 {
+	
+	double[] getUnionOfArray(double[] arr1, double[] arr2) {
+		String temp = "";
+		double[] tempArr = new double [arr1.length + arr2.length];
+		System.arraycopy(arr1, 0, tempArr, 0, arr1.length);
+		System.arraycopy(arr2, 0, tempArr, arr1.length, arr2.length);
```
harshvegada, line 26: Good point, to merging two array

harshvegada (approved):
Logic looks good to me!
Keep it up!

PR #1197, opened 2020-12-08
MOnikaS_Ass2930

MonikaSOct2020:
Assignment 29 and 30

### DIFF
--- a/src/monika/SecondMaxNum.java
+++ b/src/monika/SecondMaxNum.java
@@ -1,0 +1,30 @@
+
+package monika;
+/*Assignment - 29 :
+
+Find second maximum from array without sorting the array
+*/
+
+public class SecondMaxNum {
+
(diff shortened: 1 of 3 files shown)


src/monika/SecondMaxNum.java
```diff
@@ -0,0 +1,30 @@
...
+public class SecondMaxNum {
+
+	void MaxSecondNum(int arr[]) {
+		int first = 0;
+		int second = 0;
+		for (int i = 0; i < arr.length; i++) {
+			if (arr[i] > first) {
+				second = first;
+				first = arr[i];
+			} else if (arr[i] > second) {
```
harshvegada, line 17: you should one extra condition for max number is like

harshvegada (approved):
Rest code looks good to me!
Good to merge

PR #1211, opened 2020-12-09
RubyK_Assignment_29

rubykoct2020:
image

### DIFF
--- a/src/ruby/Assignment29.java
+++ b/src/ruby/Assignment29.java
@@ -1,0 +1,26 @@
+package ruby;
+/*/Find second Max Number/*/
+
+public class Assignment29 {
+
+	void findSecMax(int[] arr) {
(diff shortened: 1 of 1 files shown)


src/ruby/Assignment29.java
```diff
@@ -0,0 +1,26 @@
...
+/*/Find second Max Number/*/
+
+public class Assignment29 {
+
+	void findSecMax(int[] arr) {
+		int num = 0, secMax = 0;
+		for (int i = 0; i < arr.length; i++) {
+			if (num < arr[i]) {
+				num = arr[i];
+			}
```
harshvegada, line 11: you can handle both in one loop why you need to two loop?

here you need to put else if() condition like below

harshvegada (approved):
you don't require 2nd loop,
take of this next time

PR #1213, opened 2020-12-09
RubyK Assignment 31

rubykoct2020:
image


src/ruby/Assignment29.java
```diff
@@ -0,0 +1,26 @@
+package ruby;
+/*/Find second Max Number/*/
+
+public class Assignment29 {
```
harshvegada, line 4: Duplicate not in this assignment

PR #1221, opened 2020-12-10
Prasad_C_Assignment_37

PrasadCOct2020:
Hi Both,

Please review my PR.
image

Regards
Prasad Chitale

### DIFF
--- a/src/prasad/FebonacciSeries.java
+++ b/src/prasad/FebonacciSeries.java
@@ -1,0 +1,37 @@
+package prasad;
+import java.util.Scanner;
+
+public class FebonacciSeries {
+	
+	void feboSeries(int n) {
+		int temp1=0;
+		int temp2=1;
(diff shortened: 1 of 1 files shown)


src/prasad/FebonacciSeries.java
```diff
@@ -0,0 +1,37 @@
...
+		int temp1=0;
+		int temp2=1;
+		int ans=0;
+		int count=0;
+		
+		int max=n-2;
+		while(count<=max) {
+			if(count==0) {
+				System.out.print(temp1+" "+temp2+ " ");
+				count++;
```
harshvegada, line 16: As you increasing value of cnt in if & else condition both places so write after the condition
just to correct you

harshvegada (approved):
rest logic of code looks good to me!
Good to merge!
small correct no need to update, juts FYI

PR #1223, opened 2020-12-10
NagendraT_Assignment_37

NagendraTOct2020:

### DIFF
--- a/src/nagendra/logicalPrograms/FibonacciSeries.java
+++ b/src/nagendra/logicalPrograms/FibonacciSeries.java
@@ -1,0 +1,33 @@
+package nagendra.logicalPrograms;
+
+/*
+Assignment - 37 : 10th Dec'2020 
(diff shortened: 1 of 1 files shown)


src/nagendra/logicalPrograms/FibonacciSeries.java
```diff
@@ -0,0 +1,33 @@
...
+output : 0, 1, 1, 2, 3, 5, 8, 13, 21, 34
+*/
+public class FibonacciSeries {
+
+	void getFibonacciSeries(int number) {
+		int firstNumber = 0;
+		int secondNumber = 0;
+		int fibonacciNumber = 0;
+		for (int i = 0; i < number; i++) {
+			if (i < 2) {
```
harshvegada, line 15: logically you should write else condition only & print first two number in SOP directly.

harshvegada (approved):
Code looks good to me!
Good to merge

PR #1227, opened 2020-12-10
RohanM_Assignment_29

RohanMOct2020:
image
image

### DIFF
--- a/src/rohan/SecondMaxArray.java
+++ b/src/rohan/SecondMaxArray.java
@@ -1,0 +1,27 @@
+package rohan;
+
+/*
+ * Find second maximum from array without sorting the array(using sorting).
+ */
+
+public class SecondMaxArray {
+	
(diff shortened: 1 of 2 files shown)


src/rohan/SecondMaxArray.java
```diff
@@ -0,0 +1,27 @@
...
+	}
+	
+	void findSecondMax(int[] numArr) {
+		int max=numArr[0],smax=numArr[1];
+		for(int index=2;index<numArr.length;index++) {
+			if(numArr[index]>max) {
+				smax = max;
+				max=numArr[index];
+			}
+			else if(numArr[index]>smax)
```
harshvegada, line 22: you should one extra condition for max number is like

harshvegada (approved):
Rest Code looks good to me!
Good to merge

PR #1251, opened 2020-12-10
RubyK_Assignment_35

rubykoct2020:
image


harshvegada (approved):
Logic wise code looks good to me!
Keep practicing on coding

PR #1254, opened 2020-12-11
RubyK_Assignment_38

rubykoct2020:
image

### DIFF
--- a/src/ruby/Assignment38.java
+++ b/src/ruby/Assignment38.java
@@ -1,0 +1,36 @@
+package ruby;
+
+/*Assignment - 38 : 11th Dec'2020 
+Verify number is in Fibonacci series or not..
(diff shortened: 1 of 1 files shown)


src/ruby/Assignment38.java
```diff
@@ -0,0 +1,36 @@
...
+			sum = temp1 + temp2;
+			temp1 = temp2;
+			temp2 = sum;
+			index++;
+			if (sum == n) {
+				flag = true;
+			}
+			if (flag) {
+				return true;
+			}else if (n == 0 || n == 1)
```
harshvegada, line 21: Good catch condition

harshvegada (approved):
code logic looks good to me!
Good to merge

src/ruby/Assignment38.java
rubykoct2020: thanks Harsh

PR #1265, opened 2020-12-11
AnjaliS_Assignment38

AnjaliSoct2020:
image

### DIFF
--- a/src/anjaliS/IsFibonacciSeriesorNot.java
+++ b/src/anjaliS/IsFibonacciSeriesorNot.java
@@ -1,0 +1,36 @@
+package anjaliS;
+
+import java.util.Scanner;
+
+public class IsFibonacciSeriesorNot {
+	void displayIsFibonacci(int n) {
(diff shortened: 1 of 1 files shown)


src/anjaliS/IsFibonacciSeriesorNot.java
```diff
@@ -0,0 +1,36 @@
...
+		int firstnumber = 0;
+		int secondnumber = 1;
+		for (int i = 0; i < n; i++) {
+			int sum = firstnumber + secondnumber;
+			firstnumber = secondnumber;
+			secondnumber = sum;
+			if (n == sum) {
+				System.out.println(+n + " is a fibonacciSeries");
+				break;
+			} else if (sum > n) {
```
harshvegada, line 16: We should not write else condition here, think when first time you are executing it will go into else condition & break the loop,

harshvegada (approved):
rest logic looks good to me!

PR #1270, opened 2020-12-12
Suvela_assignment38

SuvelaKoct2020:


src/suvela/assignment38/FindNuminFibonacci.java
```diff
@@ -0,0 +1,36 @@
...
+		System.out.print(i + "," + j);
+		while (add < num) {
+			add = i + j;
+			System.out.print("," + add);
+			i = j;
+			j = add;
+
+		}
+		System.out.println("-----------------");
+		if (add == num)
```
harshvegada, line 22: this condition should be in while loop, think on this, let me know if you are not clear on this point.

harshvegada (approved):
Code looks good to me!
Good to merge!

PR #1322, opened 2020-12-16
MonikaS_Assignment_37

MonikaSOct2020:


src/monika/FibSeries.java
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+
+public class FibSeries {
+
+	static void printFibonacciSeries(int length) {
+		int num1 = 0;
+		int num2 = 1;
+		System.out.print(num1 + ", " + num2 + ", ");
+		for (int i = 3; i <= length; i++) {
```
harshvegada, line 16: why you started I's value from 3,
loop is independent it is just for iteration.

harshvegada: you can start it from 1 or 0 as well, bcoz you are doing sum of currentNumber & currentNumber+1,

harshvegada (changes requested):
work on comments.
acknowledge me once you accommodate changes.

harshvegada:
Code not updated,
Merging

PR #1323, opened 2020-12-16
MonikaS_Assignment_38

MonikaSOct2020:


src/monika/VefrifyFibSer.java
```diff
@@ -0,0 +1,38 @@
...
+
+	static void verifyNumberInFibonacciSeries(int number) {
+		int num1 = 0;
+		int num2 = 1;
+		System.out.print(num1 + " " + num2 + " ");
+		int sum = 0;
+		boolean flag = false;
+		int length = 0;
+		while (sum <= number) {
+			if (sum == number) {
```
harshvegada, line 16: Well handle condition
you should also need to handle 0 & 1

harshvegada (approved):
code logic looks good to me!

PR #1332, opened 2020-12-16
AnjaliS_Assignment40

AnjaliSoct2020:
image

image

### DIFF
--- a/src/anjaliS/MeetingRoom.java
+++ b/src/anjaliS/MeetingRoom.java
@@ -1,0 +1,32 @@
+package anjaliS;
+
+public class MeetingRoom {
+	String meetingRoomName;
+	int periodOfTime;
+
+	void bookMeetingRoom() {
+		meetingRoomName = "Nalanda";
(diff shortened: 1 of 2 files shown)


src/anjaliS/MeetingRoom.java
```diff
@@ -0,0 +1,32 @@
...
+	String meetingRoomName;
+	int periodOfTime;
+
+	void bookMeetingRoom() {
+		meetingRoomName = "Nalanda";
+		periodOfTime = 1;
+
+	}
+
+	void bookMeetingRoom(String meetingRoomName, int periodOfTime) {
```
harshvegada, line 13: This is not as per the expectation,

Write another method which helps to book Other then “Nalanda” meeting room by giving name of meeting room and hrs as parameter

you should restrict user to book nalanda right?
this is not simple as simple as you write, need to put some condition

harshvegada (approved):
rest code looks good to me! @AnjaliSoct2020 
take of this next time!
good to merge!

PR #1333, opened 2020-12-17
VaibhavJ_Assignment_39

VaibhavJOct2020:
Program 1 -->
image

Program 2 -->
image

Program 3 -->
image

Program 4 -->
image

### DIFF
--- a/src/vaibhav/assignment39/Addition.java
+++ b/src/vaibhav/assignment39/Addition.java
@@ -1,0 +1,32 @@
+package vaibhav.assignment39;
+
+/*Program - 4: 
+Create a class Addition and write add method for:
+a) add two integers
+b) add two decimal
+c) add two String*/
+
+
+public class Addition {
+	
+	Addition(int num1,int num2){
+		int num3 = num1 + num2;	
+		System.out.println("Additon of two integer is "+num3);
+	}
+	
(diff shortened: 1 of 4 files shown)


src/vaibhav/assignment39/Square.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class Square {
+	
+	double side;
+	
+	Square(double side){
+		this.side=side;		
+	}
+	
+	void getArea(){
```
harshvegada, line 20: as per the method name getArea() you should return the value, Just correcting logically.
if you are giving method name & that starts with get() then you should return value,
if you are giving method name as set() then return type should be void.
just FYI

harshvegada (approved):
Code looks good to me! @VaibhavJOct2020 
Good to merge!

src/vaibhav/assignment39/Addition.java
```diff
@@ -0,0 +1,32 @@
...
+/*Program - 4: 
+Create a class Addition and write add method for:
+a) add two integers
+b) add two decimal
+c) add two String*/
+
+
+public class Addition {
+	
+	Addition(int num1,int num2){
```
harshvegada, line 12: we need to create methods, not constructor

PR #1339, opened 2020-12-17
Ankita_Asgnmnt_40

AnkitaDOct2020:
image
image

### DIFF
--- a/src/ankita/assignment/scenarioBasedProgram/MeetingRoom.java
+++ b/src/ankita/assignment/scenarioBasedProgram/MeetingRoom.java
@@ -1,0 +1,40 @@
+/*Program - 2:
+Create a class MeetingRoom having method bookMeetingRoom which should provide
+below facility.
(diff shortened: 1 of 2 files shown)


src/ankita/assignment/scenarioBasedProgram/MeetingRoom.java
```diff
@@ -0,0 +1,40 @@
...
+	String room;
+	int hrs;
+
+	void bookMeetingRoom() {
+		room = "Nalanda";
+		hrs = 1;
+	}
+
+	void bookMeetingRoom(String room, int hrs) {
+		this.room = room;
```
harshvegada, line 24: I guess requirement is like, when user pass meeting room as Nalanda that time will not process request

Write another method which helps to book Other then “Nalanda” meeting room by giving name of meeting room and hrs

harshvegada (changes requested):
Correct the code in this case, restrict user to book nalanda

harshvegada:
Code merging, code not updated!

PR #1377, opened 2020-12-23
Prasad_C_Assignment_43

PrasadCOct2020:
Hi Both,

Please review my PR.

Output:-
image

Regards 
Prasad Chitale


src/prasad/ExceptionHandeling/SumOfNosFrStrg.java
```diff
@@ -0,0 +1,52 @@
...
+			String str = words[index];
+
+			try {
+				Double no = Double.parseDouble(str);
+				nos.add(no);
+			} catch (NumberFormatException ne) {
+
+			}
+		}
+
```
harshvegada, line 28: After sum you can use function
System.out.printf("%.1f", sum);

PR #1384, opened 2020-12-23
Ankita_Asgnmnt_43

AnkitaDOct2020:
image

### DIFF
--- a/src/ankita/assignment/string/SumOfDoubleNumbersFromString.java
+++ b/src/ankita/assignment/string/SumOfDoubleNumbersFromString.java
@@ -1,0 +1,33 @@
+/*Assignment : 43 [23rd Dec'2020]
+
+WAP to sum numbers from below String
+
+input: I have 5.6 years of exp in automation and 2.3 in manual testing
+output: 7.9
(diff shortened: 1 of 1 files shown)


src/ankita/assignment/string/SumOfDoubleNumbersFromString.java
```diff
@@ -0,0 +1,37 @@
...
+			String word = arr[i];
+			try {
+				if (Character.isDigit(arr[i].charAt(0))) {
+					sum = sum + Double.parseDouble(word);
+				}
+			} catch (NumberFormatException ne) {
+				System.out.println("Number Format Exception");
+			}
+		}
+		System.out.println("Output:" + sum);
```
harshvegada, line 29: this will print ans as 7.899990 right, but we want 2 digit only after decimal, so for that you can use below function
here you have two methods to handle this scenario,

1. System.out.printf("%.1f", sum);
2. String.format("%.1f",sum);

harshvegada (approved):
Code logic looks good to me!
Keep doing well!

PR #1385, opened 2020-12-23
RaghvendraT_Assignment_43

RaghvendramaniTOct2020:
image

### DIFF
--- a/src/raghvendra/Collection/FindNumberInStringAndAdd.java
+++ b/src/raghvendra/Collection/FindNumberInStringAndAdd.java
@@ -1,0 +1,31 @@
+package raghvendra.Collection;
+/*Assignment : 43 [23rd Dec'2020]
+WAP to sum numbers from below String
+input: I have 5.6 years of exp in automation and 2.3 in manual testing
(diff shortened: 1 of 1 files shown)


src/raghvendra/Collection/FindNumberInStringAndAdd.java
```diff
@@ -0,0 +1,31 @@
...
+			String ch=words[index];
+			try {
+				if(Character.isDigit(ch.charAt(0)))
+					sum+=Double.parseDouble(ch);
+			}
+			catch(NumberFormatException ne) {
+				ne.printStackTrace();
+			}
+		}
+		System.out.println("Sum of numbers in String is: "+sum);
```
harshvegada, line 23: this will print ans as 7.899990 right, but we want 2 digit only after decimal, so for that you can use below function
1. System.out.printf("%.1f", sum);
2. String.format("%.1f",sum);

harshvegada (approved):
Code logic looks good to me!
Keep it up! @RaghvendramaniTOct2020

PR #1393, opened 2020-12-24
Asgmt_39

DeepakLOct2020:
@harshvegada , please review and merge the PR

### DIFF
--- a/src/deepak/assignment39/Addition.java
+++ b/src/deepak/assignment39/Addition.java
@@ -1,0 +1,32 @@
+package deepak.assignment39;
+
+public class Addition {
+
+
+	void addInteger(int a, int b) {
+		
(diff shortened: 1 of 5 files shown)


src/deepak/assignment39/Addition.java
```diff
@@ -0,0 +1,32 @@
+package deepak.assignment39;
+
+public class Addition {
+
+
+	void addInteger(int a, int b) {
```
harshvegada, line 6: this is not overload example @DeepakLOct2020 
Overloading: method name must be same, method parameter sequence and data type must be different.

harshvegada (approved):
Please revise OOPS concepts!
Rest code looks good to me!
take care of this next time

PR #1402, opened 2020-12-24
YogitaP_Assignment_43

YogitaPOct2020:
image

### DIFF
--- a/src/yogita/SumOfDigitFromString.java
+++ b/src/yogita/SumOfDigitFromString.java
@@ -1,0 +1,35 @@
+/*Assignment_43
+ * WAP to sum numbers from below String
+ * input: I have 5.6 years of exp in automation and 2.3 in manual testing
+output: 7.9
+
+hint: Double's parse method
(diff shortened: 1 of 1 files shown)


src/yogita/SumOfDigitFromString.java
```diff
@@ -0,0 +1,35 @@
...
+			char ch = input[i].charAt(0);
+			try {
+				if (Character.isDigit(ch)) {
+					sum += Double.parseDouble(input[i]);
+				}
+			} catch (NumberFormatException ne) {
+				System.out.println("NumberFormatException Occurs");
+			}
+		}
+		System.out.println("\nSum Of Digits From String Is:" + sum);
```
harshvegada, line 27: you can also use 
System.out.println("Sum of all the numbers: "+String.format("%.2f", sum));

harshvegada (approved):
your code also work, just wanted to add new method in current knowledge.
 try to understand,

PR #1421, opened 2020-12-27
NagendraT_Assignment_44

NagendraTOct2020:


src/nagendra/logicalPrograms/FindFrequencyOfChar.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class FindFrequencyOfChar {
+
+	void getFrequencyOfCharacter(String str) {
+		String[] input = str.split(" ");
+		int count = 1;
+		HashMap<String, Integer> map = new LinkedHashMap<String, Integer>();
+		for (int index = 0; index < input.length; index++) {
+			if (map.containsKey(input[index])) {
+				map.put(input[index], count + 1);
```
harshvegada, line 22: this will work but, ideally we should get value from MAP then update value in map

harshvegada (approved):
Code logic looks good to me!
Keep doing well!
take of this next time!

PR #1432, opened 2020-12-28
RitikaG_Assignment_47

RitikaGOct2020:
image


harshvegada (approved):
Code looks good to me!
to reduce time complexity you can use sub string method.

Refer below code:

PR #1443, opened 2020-12-28
NagendraT_Assignment_47

NagendraTOct2020:


harshvegada (approved):
Very well handle @NagendraTOct2020,
you can also add one parameter for which index you want to replace.

PR #1478, opened 2020-12-30
YogitaP_Assignment_47

YogitaPOct2020:
image

### DIFF
--- a/src/yogita/ReplaceWordFromStringWithName.java
+++ b/src/yogita/ReplaceWordFromStringWithName.java
@@ -1,0 +1,37 @@
+/*
+ Assignment - 47 : 27th Dec'2020
+Replace second last occurance of "Hi" with your name.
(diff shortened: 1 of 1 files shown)


src/yogita/ReplaceWordFromStringWithName.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package yogita;
+
+public class ReplaceWordFromStringWithName {
+	String replaceWord(String str, String wordReplace) {
+		System.out.println("String is:" + str);
+		String input[] = str.split(" ");
+		int cnt = 0;
+		for (int i = input.length - 1; i >= 0; i--) {
```
harshvegada, line 16: this also good approach, 2nd approach is you can use substring method, check below code

harshvegada (approved):
Over all code looks good to me
Good to merge!

PR #1491, opened 2021-01-02
Mrunal_Assignnment_48

MrunalYOct2020:


src/mrunal/student.java
```diff
@@ -0,0 +1,42 @@
+	/*Assignment : 48
+	Student class having attribute -> studentId, studentName, studentStd.
+	A) Use comparable to sort all students based on studentName.*/
+
+package mrunal;
+
+public class student implements Comparable {
```
KrishnaKTechnocredits, line 7: you should mention Comparable<Student>

src/mrunal/student.java
```diff
@@ -0,0 +1,42 @@
...
+		
+		private int getstudentID() {
+			return studentId;
+		}
+
+		private int getstudentStd() {
+			return studentStd;
+		}
+		
+		public int compareTo(Object arg0) {
```
KrishnaKTechnocredits, line 31: If you make Comparable<Student> then you can write Student instead of Object, then you don't require to type cast to student

KrishnaKTechnocredits (approved):
take of this next time,
Code looks good to me!

PR #1541, opened 2021-01-03
NitinJ_Assignment_43

NitinJOct2020:
image


KrishnaKTechnocredits (approved):
Code logic looks good to me!
Good approach!

But your output is 7.89999999

to get 7.9 you can use below function
System.out.printf("%.2f")

take of this next time

PR #1594, opened 2021-01-04
Sadhana_Assignment_43

SadhanaPOct2020:
@harshvegada Please review my PR.


harshvegada (approved):
Code looks good to me!
Good to merge!

just FYI
If you get ans like 5.1111111
then you can use below as function
		System.out.printf("%.2f", sum);

